PR #7, opened 2023-08-24
Change email to identity

haydentherapper:
Sigstore supports more than just email, including CI workflow identity. This clarifies that we support signing on CI platforms, not just signing with an email.


laurentsimon (approved):
The original intention was to have namespaces to avoid the need to push everything in one option. If users want to use email, they can use`--email-provider` and `--email`. If they want workload identity, we'd have another name for it. 

Your changes work too.

Issue #6, opened 2023-08-23
Add GitHub Action for linting


haydentherapper:
We need to pick a linter for python.

laurentsimon:
https://github.com/rhysd/actionlint is one that works well I think

haydentherapper:
That's good for linting actions, we'll also want one for linting python files, like https://github.com/py-actions/flake8

PR #34, opened 2023-10-20
Adding linting with flake8

haydentherapper:
Fixes https://github.com/google/model-transparency/issues/6

### DIFF
--- a/model_signing/model.py
+++ b/model_signing/model.py
@@ -24,61 +24,68 @@ from sigstore.oidc import (
 )
 from sigstore_protobuf_specs.dev.sigstore.bundle.v1 import Bundle
 from sigstore.verify import (
-    VerificationMaterials,
     policy,
     Verifier,
 )
 from sigstore.verify.models import (
     VerificationMaterials,
-    VerificationResult,
 )
 
-import os, io, hashlib, base64, json
+import os
+import io
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


model_signing/model.py
```diff
@@ -24,61 +24,68 @@
 )
 from sigstore_protobuf_specs.dev.sigstore.bundle.v1 import Bundle
 from sigstore.verify import (
-    VerificationMaterials,
     policy,
     Verifier,
 )
 from sigstore.verify.models import (
     VerificationMaterials,
-    VerificationResult,
```
mihaimaruseac, line 33: I think this one is still needed, used in 

```
def verify(self, inputfn: Path, signaturefn: Path,
               ignorepaths: [Path], offline: bool) -> VerificationResult:
```

Lines 130-152

haydentherapper: We are defining our own class here - https://github.com/google/model-transparency/pull/34/files#diff-5721baa8b2b4f803a185c5388ceecb5ff321e2d0a63671db95066063b32c3b78L109

I _think_ our redefinition is the correct one to use?

mihaimaruseac: Oh, I missed those lines, thank you for the clarification. In this case, this seems right to me.

mihaimaruseac (approved):

PR #41, opened 2023-10-24
Add model signing unit test GitHub Actions

mihaimaruseac:


mihaimaruseac:
The token doesn't seem to be available when creating the PR to add the workflow, so that's why the CI fails.

https://github.com/mihaimaruseac/model-transparency/actions/runs/6628201742/job/18004764703 is the equivalent of this on my own fork (https://github.com/mihaimaruseac/model-transparency/pull/2)

laurentsimon:
Correct. We need to check how sigstore-python did it. iirc, they pre-generate some tokens every 15mn or so and put them in a bucket; then use these for the pre-submits. There's a lot of pre-submits we can add without signing zo, like unit tests for hashing, etc. Also, we will eventually need regression tests that ensure that previously-generated signatures can be verified by new versions of the verifier

mihaimaruseac:
Yeah, I tried this as a quick way to check the dependabot PRs without needing to run manually or write longer unit tests. But probably it's better to revert to writing unit tests and running those instead.

mihaimaruseac:
That was my reasoning for sending the PR. We can also run locally (test model signing locally, run SLSA on own fork), but we'd probably still need a CI for dependency resolution to prevent #43

laurentsimon:
So let's just test the installation then? Can be a pre-submit and is not throwaway work?

mihaimaruseac:
> How about updating the pip dependencies for now and work on something similar to sigstore-python towards a proper v1 release (+ unit tests).

Yeah, that's my impression after #43, we can just make a CI that tries to install the deps and if we think something might break further try to debug locally (like I just did for #37 and #32)

mihaimaruseac:
> So let's just test the installation then? Can be a pre-submit and is not throwaway work?

Yup, working on a PR for that

PR #42, opened 2023-10-24
Revert "build(deps): bump actions/setup-python from 4.7.0 to 4.7.1"

mihaimaruseac:
Reverts google/model-transparency#30

Reverting because it breaks `pip` package resolution: [working](https://github.com/mihaimaruseac/model-transparency/actions/runs/6626589695/job/17999819797), [broken](https://github.com/mihaimaruseac/model-transparency/actions/runs/6631077841/job/18013816004#step:4:177): both have 

```
Collecting etils[array-types,enp,epath,epy,etqdm,etree]==1.5.1 (from -r slsa_for_models/install/requirements.txt (line 190))
  Using cached etils-1.5.1-py3-none-any.whl (140 kB)
```

in the log, but the one after the action bump is then trying to look for 

```
Collecting etils[epath] (from array-record==0.5.0->-r slsa_for_models/install/requirements.txt (line 18))
ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
    etils[epath] from https://files.pythonhosted.org/packages/0f/6a/d2aaebacf73d5da7126c632ec0d9dc2df99cc4bbd259bad48904a034fc1b/etils-1.5.2-py3-none-any.whl (from array-record==0.5.0->-r slsa_for_models/install/requirements.txt (line 18))
```

PR #44, opened 2023-10-24
Add CI to make sure dependencies are always installable

mihaimaruseac:
Replaces #41 by only testing that dependencies can be installed (so we no longer have to manually detect them as in #37 or get to a revert like #42). Closes #41 as deprecated upon merging.

Bumps the checkout actions to latest version as a bonus

### DIFF
--- a/.github/workflows/validate_deps.yml
+++ b/.github/workflows/validate_deps.yml
@@ -1,0 +1,45 @@
+name: Validate all Python dependencies work together
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+    types: [opened, synchronize]
+
+permissions:
+  id-token: write
+
+jobs:
+  model-signing:
+    name: Test model signing dependencies
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+    - uses: actions/setup-python@65d7f2d534ac1bc67fcd62888c5f4f3d2cb2b236 # v4.7.1
+      with:
+        python-version: 3.11
+        cache: pip
+        cache-dependency-path: model_signing/install/requirements.txt
+    - name: Install dependencies
+      run: |
+        set -euo pipefail
+        python -m venv venv
+        source venv/bin/activate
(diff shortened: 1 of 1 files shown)


.github/workflows/validate_deps.yml
```diff
@@ -0,0 +1,45 @@
+name: Validate all Python dependencies work together
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+    types: [opened, synchronize]
+
+permissions:
+  id-token: write
```
haydentherapper, line 10: Is this permission still needed?

.github/workflows/validate_deps.yml
```diff
@@ -0,0 +1,45 @@
...
+        cache: pip
+        cache-dependency-path: model_signing/install/requirements.txt
+    - name: Install dependencies
+      run: |
+        set -euo pipefail
+        python -m venv venv
+        source venv/bin/activate
+        python -m pip install --require-hashes -r model_signing/install/requirements.txt
+
+  slsa-for-m:
```
haydentherapper, line 30:
```suggestion
  slsa-for-ml:
```

.github/workflows/validate_deps.yml
mihaimaruseac: No, removing them. Copy paste left over

.github/workflows/validate_deps.yml
mihaimaruseac: Thanks, fixed

.github/workflows/validate_deps.yml
```diff
@@ -6,8 +6,7 @@ on:
     branches: [main]
     types: [opened, synchronize]
 
-permissions:
-  id-token: write
+permissions: {}
```
haydentherapper, line 9: You might need read-all?

haydentherapper (approved):

mihaimaruseac:
Hmm, It seems we're still seeing the error, so #42 did not solve the issue. Will debug further, now we have this to test.

.github/workflows/validate_deps.yml
mihaimaruseac: It should work without that, since it's just reading code, doesn't need the token

PR #45, opened 2023-10-24
Bump dependencies to close #37

### LINKED ISSUE #43: Simplify updating dependencies
In #30, we bumped actions.setup-python but then this caused issues when installing the dependencies as now some hashes were missing. So we had to revert in #42.

In #37 we only have a bump in `cryptography` but that results in incompatible dependency set
```
INFO: pip is looking at multiple versions of pyopenssl to determine which version is compatible with other requirements. This could take a while.
ERROR: Cannot install -r model_signing/install/requirements.txt (line 350) and cryptography==41.0.4 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested cryptography==41.0.4
    pyopenssl 23.0.0 depends on cryptography<40 and >=38.0.0

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
```

We will probably need 2 things:

- one workflow that just tries to install the dependencies on PRs so we quickly receive signals when something breaks (at least until we get real unit tests)
- a modality to quickly update the lock files

We probably also need to further debug the #31 / #42 issue

mihaimaruseac:
Generated via

```console
[...]$ pip install pip-tools
[...]$ pip install --upgrade ${dependencies}
[...]$ pip freeze > requirements.in
[...]$ pip-compile --allow-unsafe --generate-hashes --output-file=requirements.txt requirements.in
```

Closes #37, closes #43


haydentherapper (approved):

laurentsimon (approved):
Thanks. Feel free to send another PR documenting how the file was generated (for us maintainers)

mihaimaruseac:
Will include in README.md.

PR #59, opened 2023-10-30
fix: pytest command not found

laurentsimon:
Not sure why the command is not found, especially since the correct directory is listed in the PATH.

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -29,4 +29,4 @@ jobs:
     - name: Run unit tests
       run: |
         set -euo pipefail
-        pytest .
+        ./venv/bin/pytest .


.github/workflows/unit_tests.yml
```diff
@@ -29,4 +29,4 @@ jobs:
     - name: Run unit tests
       run: |
         set -euo pipefail
-        pytest .
\ No newline at end of file
+        ./venv/bin/pytest .
```
mihaimaruseac, line 32: Oh, you need to re-enable the environment, as each `run` is in it's own shell. See https://github.com/google/model-transparency/blob/e1d31871e88b6e8c2b821fd97786c92f9a4a03f8/.github/workflows/slsa_for_ml.yml#L43

laurentsimon: what a mistake :)

laurentsimon:
@mihaimaruseac please PTAL. I fixed the problem with your suggestion

haydentherapper (approved):

PR #61, opened 2023-10-30
fix: fix unit test for test_altered_file

laurentsimon:


model_signing/serialize_test.py
```diff
@@ -259,13 +259,13 @@ def test_different_filename(self):
     # contents.
     def test_altered_file(self):
         file = "model_file"
-        file_size = 999
```
laurentsimon, line 262: this is just to speed things up

mihaimaruseac (approved):

PR #62, opened 2023-10-31
Removed the redundant use of data variable.

gautam1858:
1.     Unpack x and y directly in the loop header, making the code cleaner.
2.     Removed the redundant use of data variable.
3.     Simplified the string formatting for the print statement.


mihaimaruseac (changes requested):
Closing since the PR does not bring any useful functionality, it only adds cosmetic changes that are also brittle

Issue #57, opened 2023-10-27
Add unit tests


laurentsimon:
For each serialization:
- file:
   - symlink
   - chunk sizes
   - different results for different file content
   - different model filename returns same result
- dir:
   - same as for file
   - different results for different folder names
   - different results for different file names
   - different results for different file contents
   - different folder name returns same result
   - ignore files
   - signature file

laurentsimon:
We should also consider resolving https://github.com/google/model-transparency/pull/64#discussion_r1378198940, which suggests reformatting some of the tests. /cc @mihaimaruseac

laurentsimon:
Update: serialize_v0 tests are finished (modulo the comment above)
serialize_v1 tests are done for files. I'll work on folder tests next

PR #64, opened 2023-10-31
feat: Add tests for serialize_v0 for directory models

laurentsimon:
Part of https://github.com/google/model-transparency/issues/57

### DIFF
--- a/model_signing/serialize_test.py
+++ b/model_signing/serialize_test.py
@@ -234,8 +234,8 @@ class Test_serialize_v0:
 
         assert (r0 == r1)
 
-    # Folder serialization returns the same results for different folder names
-    # that are ignored.
+    # Folder serialization returns the same results for different folder or
+    # file names and / or file contents.
     def test_different_ignored_paths(self):
         folder = "model_dir"
         model = create_empty_folder(folder)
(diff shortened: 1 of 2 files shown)


model_signing/serialize_test.py
```diff
@@ -9,12 +9,46 @@
...
 
 # Utility functions.
-def create_folder(name: str) -> Path:
+def create_empty_folder(name: str) -> Path:
     p = os.path.join(os.getcwd(), testdata_dir, name)
     os.makedirs(p)
     return Path(p)
 
 
+def create_random_folders(name: str) -> (Path, int, [Path], [Path]):
```
mihaimaruseac, line 18: Question: why not use [`tempfile.mkdtemp`](https://docs.python.org/3/library/tempfile.html)?

model_signing/serialize_test.py
```diff
@@ -147,14 +200,183 @@ def test_altered_file(self):
             cleanup_model(altered_model)
         cleanup_model(model)
 
-    # TODO(#57): directory support.
+    # Folder serialization raises error for negativ chunk values.
```
mihaimaruseac, line 203: Typo: negative

model_signing/serialize_test.py
laurentsimon: you mea as the location of the folder? Yeah I can do that. Maybe I should update other functions that create files as well then?

model_signing/serialize_test.py
laurentsimon: good catch. Fixed.

model_signing/serialize_test.py
mihaimaruseac: Yeah, I'm thinking of having calls to `mkdtemp`/`mkstemp` to create the directories/files (and act as context managers) and then a separate utility to write random bytes to a file. Then we could have the test be something like

```python
with random_directory() as dir:
  with random_directory(dir) as dir1:
     with random_file(dir1) as f:
        generate_random_file_contents(f)  # /dir/dir1/f actually
        sign(dir)
# now we are guaranteed that the directory is removed
```

This could also be done later, if needed. I just think it will be more readable and helps with cleaning up.

mihaimaruseac (approved):

laurentsimon:
I like the idea. Let me know if I should give it a shot in this PR or crate a tracking issue for it

mihaimaruseac:
Either way works, initially I thought we were just replicating the functionality of the `tempfile` API, but we're doing more so it's ok to push for later refactoring

PR #99, opened 2024-01-10
Make the dependency pinning action create a PR

mihaimaruseac:
The dependencies have been properly updated this time (in #96 because I had 3 different archives with the same set of files, only one OS had deps properly updates).

I tested that the action runs ok if there's no need to create a PR. Also runs ok on both trigger and cron.

Added the dependencies as updated by the action

### DIFF
--- a/.github/workflows/pin_deps.yml
+++ b/.github/workflows/pin_deps.yml
@@ -92,7 +92,7 @@ jobs:
       with:
         title: "Update frozen python dependencies"
         commit-message: "Bump frozen dependencies"
-        committer: "GitHub <noreply@github.com>"
-        author: "GitHub <noreply@github.com>"
+        committer: "Mihai Maruseac (automated) <mihaimaruseac@google.com>"
+        author: "Mihai Maruseac (automated) <mihaimaruseac@google.com>"
         signoff: true
         delete-branch: true
--- a/.github/workflows/pin_deps.yml
+++ b/.github/workflows/pin_deps.yml
@@ -1,6 +1,8 @@
 name: Pin dependencies
 on:
-  workflow_dispatch
+  workflow_dispatch:
+  schedule:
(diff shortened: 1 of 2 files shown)


mihaimaruseac:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> View this [failed invocation](https://github.com/google/model-transparency/pull/99/checks?check_run_id=20365384480) of the CLA check for more information.
> 
> For the most up to date status, view the checks section at the bottom of the pull request.

This is an issue with the CLA for the bot, but since the "contribution" in the second commit is just automatically generated I think we can override the CLA check for the bot

haydentherapper:
Can you override a CLA check? I thought that was set by the org. There might be an org-level bot we could use to create those commits that's approved for the CLA? Have to check with OSPO

mihaimaruseac:
We can as Googlers, it's an escape hatch for the cases not covered by the bot (was using that multiple times in TF)

haydentherapper:
Makes sense. This is a manual override, yea? You could also use one of the codeowners' PATs.

mihaimaruseac:
Yeah, it's manual. I'll send the link to the page offline.

We could try and force the bot to act as one codeowner and use a PAT in that case.

haydentherapper:
One concern is that the PAT couldn’t be scoped to a specific set of files. Just thinking about mitigating a risk of the PAT being compromised. 

If I’m overly complicating this, I’m also good with this approach. :)

haydentherapper (approved):

laurentsimon:
re: PAT. Would fine-grained PATs help?

mihaimaruseac:
I think the issue we landed on is that GHA is the author of the PR and the CLA check Google has requires CLA approval/bypass for it

PR #103, opened 2024-01-11
Make all workflows run on Windows and fix unit tests

mihaimaruseac:
Also fix Windows tests for v1 serialization: make sure to always treat paths as if they belong to the same OS


laurentsimon (approved):
Thanks!

PR #104, opened 2024-01-12
[NFC] Spurious linting fixes and checks

mihaimaruseac:
- ensures no trailing whitespace or empty lines at end of file can be committed (and fixes existing issues)
- ensures CI doesn't run twice when it should run only once
- ensures only Windows related dependency pins have CRLF line endings, everything else is LF.
- wrapped some lines back to 80 characters


laurentsimon (approved):
Nice!

Which linter did you run? Maybe we can add it as a pre-submit and have a helper to run it locally? (not this PR, of course)

mihaimaruseac:
I added the checks to the lint GHA, but it cannot fix automically (we run in the same issue with GHA requiring CLA approval)

Issue #111, opened 2024-01-18
Manifest file


laurentsimon:
Our current code signs / serializes folders using a custom hash built using sha256. It works well but has 3 disadvantages:
1. We need to update rekor hash support (very small change for rekord zo)
2. AI frameworks that need to verify a subset of files (e.g., a tf file in a hugging face repo) won't be able to take advantage of _existing_ signatures.
3. When small files change, it's expensive to re-hash the entire model https://github.com/google/model-transparency/issues/83 

A workaround is to create and sign a dedicated "manifest" file that lists all files within a directory with their corresponding hash (similar to SHA256SM output), rather than the output of the folder serialization. With our current code, however, this creates 2 files: the manifest and a signature (on the manifest), which is bad UX-wise, especially for single-file models.

What we want is a single file. I think a solution to this problem is to use [DSEE](https://github.com/secure-systems-lab/dsse) envelope (supported by Sigstore and widely used by tools like cosign). DSSE lets us define a payload which is stored in the envelope, along the signature. This payload is the content of our manifest. We can use intoto / json as a format since it's widely adopted.

sigstore-python added support for signing intoto statements in their main branch (yet to be released), so I think we can use that. All we need is define our own predicate and its format, something like:

```json
{
   "_type": "https://in-toto.io/Statement/v1",
   "predicateType": "https://github.com/openssf/model-signing/manifest/v1"
   "predicate": []{
       {
           "path": "path/to/file1",
            "digest": {
               "sha256-p1": "...",
            }
           "bla": "...",
       },
       {
           "path": "path/to/file2",
            "digest": {
               "sha256-p1": "...",
            }
       }
   }
}
```

laurentsimon:
I've tested the signing part and it works. Next we need verification, blocked on https://github.com/sigstore/sigstore-python/issues/628

laurentsimon:
note to myself. https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md `Set of software artifacts that the attestation applies to. Each element represents a single software artifact. Each element MUST have digest set.`

We'll need to use a dummy value for subject, since the model _is_ the artifact. Note that we cannot list each file in the `subject` header because 1) the hash is unknown to rekor and 2) the files MUST be an exact match

McPatate:
Btw regarding your question in #83 :
> If users version their models, they'd only need to sign when the model actually changes.. which means they need to re-sign large files and cannot use the per-file hash anyway... Wdut?

[...]

PR #112, opened 2024-01-19
WIP feat: Use manifest

laurentsimon:
closes https://github.com/google/model-transparency/issues/111

blocked on https://github.com/sigstore/sigstore-python/issues/628

### DIFF
--- a/model_signing/_manifest.py
+++ b/model_signing/_manifest.py
@@ -1,0 +1,60 @@
+# Copyright Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing perepo_managerissions and
+# limitations under the License.
+
+from enum import Enum
+from pathlib import Path
+from typing import IO
+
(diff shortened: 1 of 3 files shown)


laurentsimon:
This is now working with https://github.com/sigstore/sigstore-python/pull/962

High-level:
clone sigstore-python
    make dev
    (source env/bin/activate)
    python3 -m pip install psutil==5.9.8
clone this repo
  run it as usual

mihaimaruseac:
This is great! Thank you!

model_signing/_manifest.py
```diff
@@ -0,0 +1,101 @@
...
+            f = {
+                "path": p.path,
+                "digest": {
+                    str(p.hashed.algorithm): p.hashed.digest.hex(),
+                },
+            }
+            files += [f]
+        stmt = (
+            dsse._StatementBuilder()
+            .subjects(
```
susperius, line 91: Shouldn't the subjects be the files? 

My mental model had the statement subjects as the artifacts we want to sign so you get `subject[n].name = path` and `subject[n].digest = file_hash`. So we would only set the `predicate_type` field as `sigstore.dev/model-transparency/manifest/v1` and that describes everything we need to know about the subjects. Furthermore, the predicate could then hold metadata like `{"hash_algorithm": "abc"}`.

PLMK if there is something I don't get here.

See https://github.com/in-toto/attestation/blob/main/spec/v1/statement.md

laurentsimon: You're correct that's how intoto works, but there is some discussion here about whether we want to use intoto or not https://github.com/sigstore/model-transparency/issues/111

For this PR, I could not pack files in the subject due to some current limitation on the hash type accepted by sigstore-python library (they only accept known hashes, but w euse a parallel hash).

mihaimaruseac:
Closing this one now that have #262 ... #267 and #276

PR #125, opened 2024-02-06
Add `tensorlfow-datasets` dependency

mihaimaruseac:
It seems that SLSA for models workflows fails when building TF models (example run:
https://github.com/mihaimaruseac/model-transparency/actions/runs/7794330107/job/21255472664). Apparently, when we added the workflow to freeze dependencies we forgot to add this dependency.


mihaimaruseac:
This also replaces #124

laurentsimon (approved):
Thanks. Can we run the workflows on PR to catch these problems (not in this PR)? We can add an `if: ...` to avoid calling the SLSA generator reusable workflow. Wdut?

PR #134, opened 2024-02-21
Bump frozen dependencies

mihaimaruseac:
Replaces #133 and solves 3 Dependabot issues in `cryptography`


mihaimaruseac:
I misread. We don't need this one as we can merge #133 instead

PR #177, opened 2024-05-15
Reimplementation and added features

susperius:
As mentioned during the meeting this the re-implementation I was working on.

It provides model signing through

- sigstore
- bring your own key (ecdsa)
- bring your own PKI (ecdsa based keys)
- no signing at all just create a bundle

I tried to design it in a way that allows easy usage of new Signer/Verifier pairs and plugin new hash algorithm. The hash algorithm part is missing a layer of abstraction to make it really easy. My hash implementation is just a naive parallel implementation based on files. It is slightly slower on my MacBook than the benchmark numbers (0.2 second for bertseq2seq and 1 second for falcon 7b).

The signing and verification are now to separate scripts. It felt way overloaded to have it in a single command so along the lines of one tool for one purpose ....

Please let me know what you think and provide feedback.

Cheers,
Martin


laurentsimon (commented):
Overall this looks really good. It aligns with the API suggested in https://github.com/sigstore/model-transparency/issues/172 (please take a look).

Could we split the PR, maybe by starting with the `hashing` folder? You can keep this PR open for reference; it's really helpful to see the overall code. But I think starting smaller will help us review faster. Wdut?

model_signing/hashing/model_hash.py
```diff
@@ -0,0 +1,92 @@
...
+        # TODO: clean path; make sure to only store parts that are in the model folder.
+        # e.g. instead of /a/b/c/model_base/weights/chunk1-1000.gz only store weights/chunk1-100.gz
+        name=internal_path,
+        digest={'sha256': __hash_file(full_path)}
+    )
+
+
+def hash_model(
+        path: str,
+        ignore: Optional[list[str]] = ['signature.json'],
```
laurentsimon, line 53: nit: shall we call this `model.sig` for folder models? This way we don't expose the format of the signature. If we ever want to change it or add a different format, we won't break existing implementations.

For file-only model, we can call it `<model-name>.sig`, like in the existing code.

Wdut?

model_signing/hashing/model_hash.py
```diff
@@ -0,0 +1,92 @@
...
+
+from concurrent import futures
+
+from in_toto_attestation.v1 import statement 
+from in_toto_attestation.v1 import resource_descriptor as res_desc
+
+
+_CHUNK_SIZE = int(1_000_000_000) # ~1GB
+_MAX_WORKERS = 10
+
```
laurentsimon, line 29: nit: let's make this file folder "private" to make sure we communicate to users they should not rely on the interface for the time being.

model_signing/hashing/model_hash.py
```diff
@@ -0,0 +1,92 @@
...
+import pathlib
+
+from concurrent import futures
+
+from in_toto_attestation.v1 import statement 
+from in_toto_attestation.v1 import resource_descriptor as res_desc
+
+
+_CHUNK_SIZE = int(1_000_000_000) # ~1GB
+_MAX_WORKERS = 10
```
laurentsimon, line 28: do we need to specify max workers? I think python has a threshold set by default, so we can let the runtime decide?

model_signing/hashing/model_hash.py
```diff
@@ -0,0 +1,92 @@
...
+        # e.g. instead of /a/b/c/model_base/weights/chunk1-1000.gz only store weights/chunk1-100.gz
+        name=internal_path,
+        digest={'sha256': __hash_file(full_path)}
+    )
+
+
+def hash_model(
+        path: str,
+        ignore: Optional[list[str]] = ['signature.json'],
+        model_metadata: dict[str, str]={'name': 'unknown'}) -> statement.Statement:
```
laurentsimon, line 54: was curious if e should make this function return bytes or something generic, so that if we ever change the implementation, it won't be a breaking change.

laurentsimon (commented):
Thanks so much for the PR. Looks really god overall.I left a few comments.

model_signing/hashing/model_hash.py
```diff
@@ -0,0 +1,92 @@
...
+        model_metadata (_type_, optional): metadata about the model. Defaults to {'name': 'unknown'}.
+
+    Raises:
+        ValueError: if the model does not exist
+        ValueError: if the input path is not a directory
+
+    Returns:
+        statement.Statement: containing the file hashes of the model and the provided metadata.
+    """
+    subjects = []
```
laurentsimon, line 69: nit: shall we make all paths absolute to simplify the rest of the code? We're doing some replace and other operations, and it's not obvious what happens if paths are relatives r don't have tailing "/", etc

model_signing/hashing/model_hash.py
```diff
@@ -0,0 +1,92 @@
...
+    if not os.path.exists(path):
+        raise ValueError(f'path "{path}" does not exist')
+    if not os.path.isdir(path):
+        raise ValueError('hash_model expects a directory as input')
+    filepaths = []
+    for dir_path, _, file_names in os.walk(path):
```
laurentsimon, line 75: how are symlinks treated?

font (commented):
@susperius Thanks for working on this! I made a first pass and it generally looks really good. I made a few comments. The main one being that I think we should also offer a way to sign and upload to Sigstore's Rekor transparency log using an existing key instead of just the keyless option. What do you think?

PR #182, opened 2024-05-16
WIP: updated benchmarking script to work with Darwin

Gregory-Pereira:
Relates to: https://github.com/sigstore/model-transparency/issues/78

#### Summary

The benchmarking script does not work with Darwin, in such a way that it appears to be suffering under load while it actually just doesn't work in a automated fashion. This PR will seek to update the benchmarking script to work for both Linux and Darwin.

#### Release Note

Use `jq` and `.json` files as results store instead of bash associative arrays

#### Documentation

No documentation should be impacted by this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+model_signing/test_env/
+model_signing/YOLOP/
+*/__pycache__/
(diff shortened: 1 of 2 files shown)


mihaimaruseac (commented):
We are likely going to drop away this benchmark file. It's not the one we used for the latest set of benchmarks

PR #184, opened 2024-05-17
Show the federated IdP in the CLI instead of the Sigstore DEX instance

### LINKED ISSUE #157: Fix IdP shown in CLI
See https://github.com/sigstore/sigstore-python/issues/970

We need to extract data from the cert ourselves to get the IdP shown to users, instead of using the Sigstore DEX instance.

Update should be this line https://github.com/sigstore/model-transparency/blob/main/model_signing/model.py#L104

mayaCostantini:
Resolves #157

#### Summary

Show the federated IdP instead of the Sigstore DEX instance, extracted from `sigstore.oidc.IdentityToken`.
As mentioned in the comment, this is more of a workaround for versions of `sigstore-python` earlier than `3.0.0`, as the `federated_issuer` property could be called directly in the latest version.

#### Release Note

Replaced DEX instance with federated IdP in signing CLI field `identity-provider`.

#### Documentation

No documentation required.

### DIFF
--- a/model_signing/model.py
+++ b/model_signing/model.py
@@ -101,7 +101,10 @@ class SigstoreSigner():
             oidc_token = self.get_identity_token()
             if not oidc_token:
                 raise ValueError("No identity token supplied or detected!")
-            print(f"identity-provider: {oidc_token.issuer}",
+            # Calling the private attribute IdentityToken._federated issuer
+            # is a workaround for earlier versions of sigstore-python (<3.0.0)
+            # that do not support the federated_issuer property.
+            print(f"identity-provider: {oidc_token._federated_issuer}",
                   file=sys.stderr)
             print(f"identity: {oidc_token.identity}", file=sys.stderr)
 


mihaimaruseac (dismissed):
Thank you!

mihaimaruseac:
Oh, can you fix the lint issues, please (lines too long)

model_signing/model.py
```diff
@@ -101,7 +101,10 @@ def sign(self, inputfn: Path, signaturefn: Path,
             oidc_token = self.get_identity_token()
             if not oidc_token:
                 raise ValueError("No identity token supplied or detected!")
-            print(f"identity-provider: {oidc_token.issuer}",
+            # Calling the private attribute IdentityToken._federated issuer is a workaround
+            # For earlier versions of sigstore-python (<3.0.0) that do not support the
+            # federated_issuer property.
+            print(f"identity-provider: {oidc_token._federated_issuer}",
```
haydentherapper, line 107: Is federated issuer always set? If a non-Dex token is used, what is this value?

laurentsimon: @woodruffw please advise

laurentsimon: @mayaCostantini this is the original issue + solution that @woodruffw told us should work https://github.com/sigstore/sigstore-python/issues/970#issuecomment-2108727868. So the right solution is probably `.expected_certificate_subject`

laurentsimon (commented):
Thanks for the PR!

model_signing/model.py
woodruffw: With `sigstore-python` 3.0.0, `federated_issuer` should fall back to the "top-level" issuer if the Dex-specific federation claim isn't present. You can see that here:

https://github.com/sigstore/sigstore-python/blob/29c98ea7f58c3047b9b5e062af912f4523ff1514/sigstore/oidc.py#L208-L227

(Before 3.0.0, `federated_issuer` was called `expected_certificate_subject`, which was a misleading name. But they're the same API.)

haydentherapper:
Can you fix the lint failure? LGTM otherwise, thanks!

mayaCostantini:
Thanks for the reviews, the linting issues should be fixed now.

mihaimaruseac (approved):

mihaimaruseac:
Thank you!

Issue #172, opened 2024-05-13
APIs design


laurentsimon:
This issues proposes what the (long-term) APIs will look like. Looking for comments, nothing set in stone.

### 1. Hash engine
Tracked in https://github.com/sigstore/model-transparency/issues/140
Why: 
- for callers to customize the hash parameters
- for callers to customize hash implementation, eg using hw-accelerattion or distributed across machines.

```python
class hash_engine:
  def name() -> str:
  # Shard size
  def shard() -> int
  # Chunk size
  def chunk() -> int
  # Standard hash APIs
  def update(...):..
  def final():..
```

We will provide  default hash engine that we use in this library, with the possibility to customize its parameters:
```python
class shah256pv1:
   def __init__(self, shard=1000000, chunk=memory_available())
   # All class function from hash_engine
```

A hash name is parameterized. I suggest something simple like `<name>$param1$param2...`. For the existing sha256p, it could be `sha256pv1$1000000` for a shard of 1 GB.

### 2. Serializer
This will serialize a model (folder or file).
Why: some callers may want to serialize models using our library but not sign it with our library.

```python
class serializer:
   def __init__(self, hash: hash_engine)
      ...
   def run(
          # Path is a file-system path.
          # PathIter is a path iterator, for callers who want to keep things in memory.
          input: Path | PathIter,
          # A list of paths to recompute Tracked in https://github.com/sigstore/model-transparency/issues/160
          recompute_paths: []Path = None,
          # HashedPathIter is an iterator for path+hash computed
          ) -> HashedPathIter:
       
     ...
)
```

### 3. Raw Signer / Verifier
This provides low-level signer: It allows to abstract away how private key is managed (TUF, HSM, static key, etc). The raw verifier lets callers set a custom verifier if needed, eg if the signature uses a different encoding, etc. 

```python
class RawSigner:
  @abstract
  def sign(raw: bytes) -> bytes
     # Perform signature via HSM API, use PCKS11 API, use raw bytes from memory, etc
``` 
We can have helper function for common raw signers like (one will be the default?):

```python
class ECDSARawSigner(RawSigner):
  def from_path(path: filepath.Path):
     ...
  def from_bytes(key: bytes):
     # Verifies that 1) the key is of the right type, eg right length and on the right curve, 2) private and public key correspond to the same key pair. 
    ...
  def sign(raw):
   # Use the raw key
   ...
```

### 4. Signer / Verifier
[...]

Issue #140, opened 2024-03-05
Flexible hash engine


laurentsimon:
We may want to:
1. Support hashing for GPU
2. Have an API that allows passing in a custom hash engine. Callers can take advantage of their own hardware to speed up hashing. Alternatively, allow passing the hash as input

mihaimaruseac:
I think 2 is done via #188 and GPU support should not be as part of v1 release. Ok to close and open a new feature request for GPU hashing? Or leave this as is but just exclude from milestone?

laurentsimon:
LGTM

mihaimaruseac:
Made #201 for the GPU case and can close this now

Issue #13, opened 2023-08-25
Investigate different hashes


laurentsimon:
E.g. https://www.blake2.net (reduced rounds), https://github.com/BLAKE3-team/BLAKE3

laurentsimon:
BLAKE3 may be similar to what we've implemented - see comments in https://github.com/sigstore/sigstore-python/issues/1018#issuecomment-2111322094

laurentsimon:
There seems to be a package available https://pypi.org/project/blake3/

laurentsimon:
One thing to be aware of is that blake2 is not a NIST hash, and blake3 seems based on a similar idea (reduced number of rounds in the compression function). It also fixes the shard size to 1KB, which we would likely want to be parameterizable in practice.

smeiklej:
When we benchmarked model signing for different model sizes, disk performance became the bottleneck quite quickly, so my sense is that changing the hash function would have little to no impact.  
(Graph is here if you're curious: [signing_runtime.pdf](https://github.com/sigstore/model-transparency/files/15337650/signing_runtime.pdf).)

That being said, supporting different hash functions is good for the sake of modularity.  But that seems like a different goal from what you're talking about here.

laurentsimon:
How do you know disk performance was the bottleneck? It could be that the number of vCPUs was the bottleneck, since we parallelize by dispatching computation on each vCPU. Is there another graph showing that disk IO was the bottleneck?

smeiklej:
That's fair, we don't know for sure (for reference though the machine had 48 vCPUs).  But even if the bottleneck is the number of vCPUs I still don't see why a different hash function would affect performance?

laurentsimon:
If the hash can compute faster each shard on each vCPU, we'd get a lower plateau (first part of graph) and a slower linear increase (second part of graph). Of course, memory also comes into play and could become the bottleneck too :)
Wdut?

smeiklej:
One thing I forgot to mention: You can see on the graph that the results are very similar across M2 and M3 (and at the end identical), despite the fact that M3 has twice the number of cores.  So this again suggests that the bottleneck is disk IO.

@mihaimaruseac has a graph about memory usage showing that it really never got that high (like 4-8 GB RAM if I remember correctly) so I think that unless we seriously increase the shard size this is unlikely to become a bottleneck.

mihaimaruseac:
[...]

PR #188, opened 2024-05-22
Add hashing API for model signing

mihaimaruseac:
#### Summary
This is the lowest layer of the model signing API (#172). It only supports computing the digest of a single object, in a flexible way (#140). We add a precomputed hasher to allow benchmarking (remove startup, etc. time; this will come up later). We add one memory hasher for now, but we might add others later (#13).

The important part of the API is under `hashing/file`. We have 2 ways to hash a file: either completely or by passing a span (start, end pair) and only hashing the contents within the span. Next level API will generate corresponding spans to hash a file in a distributed fashion. This is also useful for models that are loaded in a distributed way: each host is able to hash only the part that it accesses.

Tested added in this CL, microbenchmarks will follow later.

Drive-by-fixes:
- add a missing `.gitignore`
- add a missing copyright notice and fix some typos in the existing ones (separated to #189 for convenience)
- fix `serialize_test.py` import to work with directory packages. We will remove this later once the API is implemented and code migrated, with testing in the proper places
- increase `flake8` max line length to 80 to match Google style

#### Release Note
NONE

#### Documentation
NONE


mihaimaruseac:
Don't merge this, I'm also doing an internal review and didn't set up Copybara yet :)

mihaimaruseac:
This is now ready for review.

model_signing/hashing/file.py
```diff
@@ -0,0 +1,223 @@
...
+
+        digest = self._content_hasher.compute()
+        return hashing.Digest(self.digest_name, digest.digest_value)
+
+    @override
+    @property
+    def digest_name(self) -> str:
+        if self._digest_name_override is not None:
+            return self._digest_name_override
+        return f"file-{self._content_hasher.digest_name}-{self._shard_size}"
```
laurentsimon, line 223: curious if we need the prepended 'file-' string or not?

laurentsimon (approved):
Thanks!

model_signing/hashing/file.py
mihaimaruseac: We need it, otherwise we won't be able to differentiate between an instance of the `_content_hasher` and an instance of the wrapped `FileHasher` (not the sharded one).

But we have the option to override this for hashes that we want to export in manifests

PR #190, opened 2024-05-28
Migrate `serialize_v0` to new API (as part of serialization layer)

mihaimaruseac:
#### Summary
Migrate `serialize_v0` to new API.
    
This is the middle layer of the API design work (#172). We add a manifest abstract class to represent various manifests (#111 #112) and also ways to serialize a model directory into manifests and ways to verify the manifests.
    
For now, this only does what was formerly known as `serialize_v0`. The v1 and the manifest versions will come soon.
    
Note: This has a lot of inspiration from #112, but makes the API work with all the usecases we need to consider right now.

#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/model_signing/manifest/manifest.py
+++ b/model_signing/manifest/manifest.py
@@ -1,0 +1,39 @@
+# Copyright 2024 The Sigstore Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Machinery for representing a serialized representation of an ML model.
+
+Currently, we only support a manifest that wraps around a digest. But, to
+support incremental updates and partial signature verification, we need a
+manifest that lists files and their digests. That will come in a future change,
+soon.
+"""
+
+from abc import ABCMeta
+from dataclasses import dataclass
+
+from model_signing.hashing import hashing
+
+
+class Manifest(metaclass=ABCMeta):
+    """Generic manifest file to represent a model."""
+
+    pass
+
+
+@dataclass
+class DigestManifest(Manifest):
+    """A manifest that is just a hash."""
+
+    digest: hashing.Digest
--- a/model_signing/serializing/dfs.py
+++ b/model_signing/serializing/dfs.py
@@ -14,6 +14,7 @@
 
 """Model serializers that build a single hash out of a DFS traversal."""
 
+import base64
 import pathlib
 from typing import Callable
 from typing_extensions import override
@@ -44,7 +45,8 @@ def _build_header(*, entry_name: str, entry_type: str) -> bytes:
         entry_type: The type of the entry (file or directory).
     """
     encoded_type = entry_type.encode("utf-8")
-    encoded_name = entry_name.encode("utf-8")
+    # Prevent confusion if name has a "." inside by encoding to base64.
+    encoded_name = base64.b64encode(entry_name.encode("utf-8"))
     # Note: make sure to end with a ".".
     return b".".join([encoded_type, encoded_name, b""])
 
--- a/model_signing/serializing/dfs.py
+++ b/model_signing/serializing/dfs.py
@@ -25,7 +25,14 @@ from model_signing.serializing import serializing
 
 
 def _check_file_or_directory(path: pathlib.Path) -> bool:
-    """Checks that the given path is either a file or a directory."""
+    """Checks that the given path is either a file or a directory.
+
+    There is no support for sockets, pipes, or any other operating system
+    concept abstracted as a file.
+
+    Furthermore, this would return False if the path is a broken symlink, if it
+    doesn't exists or if there are permission errors.
+    """
     return path.is_file() or path.is_dir()
 
 
@@ -38,6 +45,7 @@ def _build_header(*, entry_name: str, entry_type: str) -> bytes:
     """
     encoded_type = entry_type.encode("utf-8")
     encoded_name = entry_name.encode("utf-8")
+    # Note: make sure to end with a ".".
     return b".".join([encoded_type, encoded_name, b""])
 
 
@@ -68,10 +76,12 @@ class DFSSerializer(serializing.Serializer):
 
     @override
     def serialize(self, model_path: pathlib.Path) -> manifest.Manifest:
-        # TODO(mihaimaruseac): Add checks for symlinks
+        # TODO(mihaimaruseac): Add checks to exclude symlinks if desired
         if not _check_file_or_directory(model_path):
             raise ValueError(
-                f"Must have a file or directory, but '{model_path}' is neither."
+                f"Cannot use '{model_path}' as file or directory. It could be a"
+                " special file, it could be missing, or there might be a"
+                " permission issue."
             )
 
         if model_path.is_file():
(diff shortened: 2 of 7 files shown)


model_signing/manifest/manifest.py
```diff
@@ -0,0 +1,39 @@
...
+"""
+
+from abc import ABCMeta
+from dataclasses import dataclass
+
+from model_signing.hashing import hashing
+
+
+class Manifest(metaclass=ABCMeta):
+    """Generic manifest file to represent a model."""
```
laurentsimon, line 30: Will it possible to add `payload()` and `payload_type()` to support DSSE? How would that work with the simple `DigestManifest` class? Will we need some additional `bundleFormat() -> DSSE | StandaloneSignature` function?

(Need not address this in this PR)

model_signing/serializing/dfs.py
```diff
@@ -0,0 +1,118 @@
...
+def _check_file_or_directory(path: pathlib.Path) -> bool:
+    """Checks that the given path is either a file or a directory.
+
+    There is no support for sockets, pipes, or any other operating system
+    concept abstracted as a file.
+
+    Furthermore, this would return False if the path is a broken symlink, if it
+    doesn't exists or if there are permission errors.
+    """
+    return path.is_file() or path.is_dir()
```
laurentsimon, line 36: what happens on symlinks? Are they considered files and do we need to filter them out? One worry with symlinks is that they may enable writing arbitrary files via dir traversals

model_signing/serializing/dfs.py
```diff
@@ -0,0 +1,118 @@
...
+
+def _build_header(*, entry_name: str, entry_type: str) -> bytes:
+    """Builds a header to encode a path with given name and type.
+
+    Args:
+        entry_name: The name of the entry to build the header for.
+        entry_type: The type of the entry (file or directory).
+    """
+    encoded_type = entry_type.encode("utf-8")
+    encoded_name = entry_name.encode("utf-8")
```
laurentsimon, line 47: what if the encoded file name contains a `.`? We may need to encode the name, eg via base64 to avoid this issue

model_signing/serializing/dfs.py
```diff
@@ -0,0 +1,118 @@
...
+            merge_hasher_factory: A callable that returns a
+              `hashing.StreamingHashEngine` instance used to merge individual
+              file digests to compute an aggregate digest.
+        """
+        self._file_hasher = file_hasher
+        self._merge_hasher_factory = merge_hasher_factory
+
+    @override
+    def serialize(self, model_path: pathlib.Path) -> manifest.Manifest:
+        # TODO(mihaimaruseac): Add checks to exclude symlinks if desired
```
laurentsimon, line 79: create a tracking issue?

model_signing/serializing/dfs.py
```diff
@@ -0,0 +1,118 @@
...
+        if not _check_file_or_directory(model_path):
```
laurentsimon, line 80: shall we move the error raise inside `_check_file_or_directory()`? This way we won't need to keep the error message in sync with what the function does internally

model_signing/serializing/dfs.py
```diff
@@ -0,0 +1,118 @@
...
+            )
+
+        if model_path.is_file():
+            self._file_hasher.set_file(model_path)
+            return manifest.DigestManifest(self._file_hasher.compute())
+
+        return manifest.DigestManifest(self._dfs(model_path))
+
+    def _dfs(self, directory: pathlib.Path) -> hashing.Digest:
+        # TODO(mihaimaruseac): Add support for excluded files
```
laurentsimon, line 94: create an issue if not addressed in this PR

model_signing/serializing/dfs_test.py
```diff
@@ -0,0 +1,282 @@
...
+        serializer = dfs.DFSSerializer(file_hasher, memory.SHA256)
+        manifest = serializer.serialize(sample_model_file)
+
+        new_name = sample_model_file.with_name("new-file")
+        new_file = sample_model_file.rename(new_name)
+        new_manifest = serializer.serialize(new_file)
+
+        assert manifest == new_manifest
+
+    def test_file_model_hash_changes_if_content_changes(
```
laurentsimon, line 114: it may be a good idea to be more thorough how we update models: iterate over each character and each file name to make sure we have everything covered.

model_signing/serializing/dfs_test.py
```diff
@@ -0,0 +1,282 @@
...
+    def test_empty_folder(self, empty_model_folder):
+        file_hasher = file.FileHasher("unused", memory.SHA256())
+        serializer = dfs.DFSSerializer(file_hasher, memory.SHA256)
+        manifest = serializer.serialize(empty_model_folder)
+        expected = (
+            "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
+        )
+        assert manifest.digest.digest_hex == expected
+
+    def test_empty_folder_hashes_the_same_as_empty_file(
```
laurentsimon, line 193: maybe add an additional test that changes a non-empty file to a dir with the same name

model_signing/manifest/manifest.py
mihaimaruseac: Yes, I excluded them for this PR, but will do so when we get to the manifest (v2) serialization

model_signing/serializing/dfs.py
mihaimaruseac: This returns true here, but I added a TODO to handle them later

model_signing/serializing/dfs.py
mihaimaruseac: Good idea. Made #196

model_signing/serializing/dfs.py
mihaimaruseac: Good idea. Made #196

model_signing/serializing/dfs_test.py
mihaimaruseac: I'm not sure. It will slow down testing and the only thing it can catch is if any hashing engine has collisions.

model_signing/serializing/dfs_test.py
mihaimaruseac: Good idea. Added.

model_signing/serializing/dfs.py
laurentsimon: nit: use `TODO(196)` to link to the issue

model_signing/serializing/dfs.py
laurentsimon: nit, use `TODO(196)` in comment

model_signing/serializing/dfs_test.py
laurentsimon: it may uncover bugs for more complicated serializations, eg with sharding and different parameters. Here you're right it's simple enough that we're probably OK. Note that we could add just a couple test function and small models to minimize performance during testing.

laurentsimon (approved):

model_signing/serializing/dfs.py
mihaimaruseac: I'll update on the next one

model_signing/serializing/dfs_test.py
mihaimaruseac: Let's do it for end-to-end testing (#5), added a comment there to refer to this

mihaimaruseac:
Currently blocked by DCO (https://github.com/dcoapp/app/issues/211). Will make new PRs on top of this one and rebase when this gets merged.

model_signing/manifest/manifest.py
```diff
@@ -0,0 +1,39 @@
...
+
+    pass
+
+
+@dataclass
+class DigestManifest(Manifest):
```
susperius, line 36: How does a model look like as a manifest?
As the digest manifest is defined now, it contains a single digest. Shouldn't it contain a list of digests to represent the model?

Where would we place the signature in this manifest? 

Perhaps we could align a bit more with the sigstore bundle type and least provide a model manifest that contains the digests, the signature and verification material.

mihaimaruseac: Working on that now, hope to have a PR later today/early tomorrow. This PR and the one after it only migrated the old `serialize_v0`/`serialize_v1` versions

PR #194, opened 2024-06-03
Add support for hashing files with header.

mihaimaruseac:
#### Summary
Missed this in #188, but found out I need it when working on #190. The `serialize_v0`/`serialize_v1` methods all had headers in front of the files, so we need to do that too. Will update usage of header on #190 shortly.

As a benefit, we can simulate hashing a file with a header for the first portion of the file and a sharded hasher for the remainder of the file.

#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/model_signing/hashing/file.py
+++ b/model_signing/hashing/file.py
@@ -28,10 +28,30 @@ Example usage for `ShardedFileHasher`, reading only the second part of a file:
 ```python
 >>> with open("/tmp/file", "w") as f:
 ...     f.write("0123abcd")
->>> hasher = ShardedFileHasher("/tmo/file", SHA256(), start=4, end=8)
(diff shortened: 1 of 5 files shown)


laurentsimon (commented):
It's a bit dangerous that:
"header" + "file_content" will return the same hash as "head" + "erfile_content"

What's the use case?

mihaimaruseac:
Actually, I don't need it.

In the old serialization we had

```py
        h = hashlib.sha256(header)
        with open(path, "rb") as f:
            if chunk == 0:
                all_data = f.read()
                h.update(all_data)
            else:
                # Compute the hash by reading chunk bytes at a time.
                while True:
                    chunk_data = f.read(chunk)
                    if not chunk_data:
                        break
                    h.update(chunk_data)
        return h.digest()
```

So we needed a header before hashing the file. But we actually have 2 different hashers here, so this is not needed.

PR #198, opened 2024-06-05
Add API for sharded serialization to a digest

mihaimaruseac:
#### Summary
This is what used to be `serialize_v1`.

Additionally, in this change we rename `serializing` to `serialization` to be gramatically correct. We expose `shard_size` and a new `digest_size` method from all hashing engines. We also make imports be more consistent.

#### Release Note
NONE

#### Documentation
NONE


mihaimaruseac:
Lint would be fixed after #199

model_signing/serialization/dfs.py
```diff
@@ -0,0 +1,303 @@
...
+    encoded_name = base64.b64encode(entry_name.encode("utf-8"))
+
+    if start is not None and end is not None:
+        # Note: make sure to end with a ".".
+        encoded_range = f"{start}-{end}.".encode("utf-8")
+    else:
+        # Note: no "." at end here, it will be added by `join` on return.
+        encoded_range = b""
+
+    return b".".join([encoded_type, encoded_name, encoded_range])
```
laurentsimon, line 70: this will output `type.name.xxx-yyy`. I think we need an additional `.` at the end of the string, otherwise it's malleable `type.name.xxx-yyy + content == type.name.xxx-y + yycontent`. I think we're doing plain concatenation of header and conntent. Lmk if that's not the case

model_signing/serialization/dfs.py
```diff
@@ -0,0 +1,303 @@
...
+
+    @override
+    def serialize(self, model_path: pathlib.Path) -> manifest.Manifest:
+        # TODO: #196 - Add checks to exclude symlinks if desired
+        _check_file_or_directory(model_path)
+
+        if model_path.is_file():
+            entries = [model_path]
+        else:
+            # TODO: When Python3.12 is the minimum supported version, this can
```
laurentsimon, line 198: nit: create a tracking issue

model_signing/hashing/file.py
```diff
@@ -219,4 +225,4 @@ def compute(self) -> hashing.Digest:
     def digest_name(self) -> str:
         if self._digest_name_override is not None:
             return self._digest_name_override
-        return f"file-{self._content_hasher.digest_name}-{self._shard_size}"
+        return f"file-{self._content_hasher.digest_name}-{self.shard_size}"
```
laurentsimon, line 228: note: here we have to be careful that digest name do not contain the character `-`. Either we add a check somewhere, or we use a different character, like `$`, `|`. Maybe you're thinking of another PR to customize this value anyway, since `name-shard_size` could lead to confusion anyway since the sharding "algo" is not specified

model_signing/serialization/dfs.py
```diff
@@ -0,0 +1,303 @@
...
+            # be replaced with `pathlib.Path.walk` for a clearer interface.
+            entries = sorted(model_path.glob("**/*"))
+
+        sized_paths = self._get_sizes(entries, model_path)
+        tasks = self._build_tasks(sized_paths)
+
+        digest_len = self._merge_hasher.digest_size
+        digests_buffer = bytearray(len(tasks) * digest_len)
+
+        with concurrent.futures.ThreadPoolExecutor(
```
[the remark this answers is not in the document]
mihaimaruseac: We cannot use `ProcessPoolExecutor` in a library. It's supposed to have visibility into `__main__`.

Also, here we are actually better with using threads, since the computation is IO bound, not CPU. Each file read releases the GIL.

I'll run a benchmark with both, of course.

model_signing/hashing/file.py
mihaimaruseac: Hmm, we could standardize on using `.` everywhere between the fields.

So far, we control all the components so there's no risk of confusion. I'm pushing testing /more customization for this after we have most of the API in place (at least after we have the full manifest format), as we should be able to quickly fix issues.

model_signing/serialization/dfs.py
mihaimaruseac: I'm adding a dot at the end in line 65 (`{start}-{end}.`)

model_signing/serialization/dfs.py
mihaimaruseac: Good idea, but it won't be for the v1 milestone. Made #200

model_signing/serialization/dfs.py
laurentsimon: might be more readable to add to the `return` line... but does not work with the empty encoded_range :/

laurentsimon (approved):

PR #217, opened 2024-06-21
Add support for BLAKE2 hashing.

mihaimaruseac:
#### Summary
Add support for BLAKE2 hashing via the `hashlib` library. Someone else can also add support for BLAKE3 if needed, this was to test that the hashing layer really works with various hashing engines.

Next, on this stream, I'm planning on adding benchmarks (via google-benchmarks for micro benchmarking and hyperfine for large scale benchmarking). This way we can analyze the slowdown in large benchmarks and determine if it is caused by the hashing function or the operating system.

But I'll focus on the manifest for the next few PRs, currently just splitting up parallel work.

#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/model_signing/hashing/memory.py
+++ b/model_signing/hashing/memory.py
@@ -67,5 +67,38 @@ class SHA256(hashing.StreamingHashEngine):
     @override
     @property
     def digest_size(self) -> int:
-        """The size, in bytes, of the digests produced by the engine."""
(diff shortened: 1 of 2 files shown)


mihaimaruseac:
(this was mentioned in #13)

laurentsimon (approved):

PR #219, opened 2024-06-21
Prepare for manifest

mihaimaruseac:
#### Summary
The important part of the PR is the exporting of `_endpoints` and `_check_file_or_directory` as I'm using them to build the manifest that contains every file digest instead of just one single digest.

While doing these transformations, I also cleaned some typos, made comments be correct instead of copy-pasted from other places. Made sure docstrings document all the arguments and yield/return types. Changed TODOs to fully link to the GitHub issue instead of just the number. And refactored 2 functions names to be clearer on their purpose (per internal review).

This is on top of #218, since I needed some changes in the test setup too. But I separated the test changes to a separate PR to make it easier to review. Speaking of, each commit here can also be reviewed individually.

#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/model_signing/serialization/dfs.py
+++ b/model_signing/serialization/dfs.py
@@ -140,7 +140,7 @@ class DFSSerializer(serialization.Serializer):
         return hasher.compute()
 
 
-def _endpoints(step: int, end: int) -> Iterable[int]:
+def endpoints(step: int, end: int) -> Iterable[int]:
     """Yields numbers from `step` to `end` inclusive, spaced by `step`.
 
     Last value is always equal to `end`, even when `end` is not a multiple of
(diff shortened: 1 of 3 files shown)


model_signing/serialization/dfs.py
```diff
@@ -188,18 +205,25 @@ def __init__(
 
     @override
     def serialize(self, model_path: pathlib.Path) -> manifest.Manifest:
-        # TODO: #196 - Add checks to exclude symlinks if desired
-        _check_file_or_directory(model_path)
+        # Note: This function currently uses `pathlib.Path.glob` so the DFS
```
laurentsimon, line 208: I think we use `sorted()` to sort https://github.com/sigstore/model-transparency/blob/db894e203611eb35027314464d66991c66bfecfc/model_signing/serialization/dfs.py#L129 and https://github.com/sigstore/model-transparency/blob/db894e203611eb35027314464d66991c66bfecfc/model_signing/serialization/dfs.py#L224 so I don't think we rely on the glob implementation?

laurentsimon (approved):

model_signing/serialization/dfs.py
mihaimaruseac: Yes, we sort the output. But we still need to treat files and directories afterwards, whereas `Path.walk` will give them to us directly and in a guaranteed DFS traversal no matter what.

The difference would arrive when we want to ignore a directory. With `glob`, we have to analyze any element in the returned list and remove if directory is a parent. With `walk`, as soon as the directory is found, we cut the search short and continue from the next entry.

PR #221, opened 2024-06-21
Add pytype as type checking linter

### LINKED ISSUE #216: Add static type checking workflow
I recently set up the project and my local environment (VS Code + Python plugin which uses Pylance as a LSP), and noticed 30+ type errors for `model_signing` based on the type hints currently there. It seems fairly common for projects with type hints to have an associated CI linting job.

There are multiple static type checkers, so we'd need to pick one:
* [mypy](https://github.com/python/mypy)
* [pyright](https://github.com/microsoft/pyright)

I was thinking mypy just to match other sigstore repos which use it:
https://github.com/sigstore/sigstore-python/blob/c5a02012fb01c0039bf45b835020806637e0bd73/CONTRIBUTING.md?plain=1#L44-L49
https://github.com/sigstore/gh-action-sigstore-python/blob/94661007ff419d4795b935732494905162e79738/Makefile#L14-L19

spencerschrock:
#### Summary

This PR adds `pytype` a dev dependency for model_signing to allow type checking. The intention is to have this as a CI step to catch new typing issues. The lint only checks the newer packages, and explicitly ignores the older `{main,model,serialize}.py` There were two fixes for the packages and modules which are being checked (`model_signing/{hashing,manifest,serialization}`):

- `DFSSerializer.serialize` and `ShardedDFSSerializer.serialize` return `manifest.DigestManifest` instead of `manifest.Manifest`
  - The tests in `dfs_test.py` were referencing `manifest.digest` which only exists for `DigestManifest`
- Create a new `pathlib.Path` constant for `"unused"`, as `str` did not match `pathlib.Path` 

Closes #216 

#### Release Note

NONE

#### Documentation

Made a developer section in `model_signing/README.md`


spencerschrock:
Let me know about the `DFSSerializer` `ShardedDFSSerializer` changes. A `DigestManifest` is still a `Manifest`, so I assume overriding the abstract `Serializer.serialize` which returns `Manifest` isn't a problem.

mihaimaruseac (approved):
This looks good to me, thank you!

PR #230, opened 2024-07-09
Add initial support for itemized model manifests

mihaimaruseac:
#### Summary
Adds machinery for manifests that list components of the model paired with their digests. Currently supporting per file and per file shard listings, to migrate existing `serialize_v0` and `serialize_v1` functionality.

For now, we just store the manifest in memory and we don't yet support signing it. There is only one (long) comment on how all of this is planned to work, future PRs should arrive with the next functionality.

A diagram of all the classes so far:

image

Next steps involve adding the ability to serialize the manifest to disk and sign it. Then, parsing a signed manifest from disk to verify models. Then the streaming APIs. And, of course, some benchmarks, cleanup, etc., that comes in parallel.

#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/model_signing/manifest/manifest.py
+++ b/model_signing/manifest/manifest.py
@@ -25,7 +25,7 @@ could be a signature over a single digest).
 When testing the integrity of a model, we start with the model path and model
 type. From this, we extract the path to the signed manifest. Next step is to
 verify the authenticity of the signature, after which we extract an in-memory
-representation of the _trusted_ manifest. To verify the model in its integrity,
+representation of the _trusted_ manifest. To verify the integrity of the model,
 we run another `serialization.Serializer` instance to compute the _untrusted_
 manifest of the model as we're seeing it. If these 2 manifests are conforming,
 then the model integrity is maintained.
--- a/model_signing/manifest/manifest.py
+++ b/model_signing/manifest/manifest.py
@@ -28,7 +28,7 @@ manifest of the model as we're seeing it. If these two manifests agree then the
 model integrity is maintained.
 
 In the simplest case, we are working with `DigestManifest` objects. Here, the
-two manifests are conforming iff the digests are the same. A more complex case
+two manifests agree if and only if the digests are the same. A more complex case
 is when the manifest itemizes each model component (e.g., every file (or file
 shard) is paired with its digest). The manifests agree if every file and digest
 matches. Alternatively, we could allow for file renames, and check only the
--- a/model_signing/manifest/manifest.py
+++ b/model_signing/manifest/manifest.py
@@ -96,7 +96,7 @@ class ManifestItem(metaclass=abc.ABCMeta):
     pass
 
 
-@dataclasses.dataclass(frozen=True)
+@dataclasses.dataclass
 class FileManifestItem(ManifestItem):
     """A manifest item that records a filename path together with its digest.
 
@@ -105,6 +105,17 @@ class FileManifestItem(ManifestItem):
 
     path: pathlib.PurePath
     digest: hashing.Digest
+
+    def __init__(self, *, path: pathlib.PurePath, digest: hashing.Digest):
+        """Builds a manifest item pairing a file with its digest.
+
+        Args:
+            path: The path to the file, relative to the model root.
+            digest: The digest of the file.
+        """
+        # Note: we need to force a PosixPath to canonicalize the manifest
+        self.path = pathlib.PurePosixPath(path)
+        self.digest = digest
 
 
 class FileLevelManifest(ItemizedManifest):
@@ -122,7 +133,7 @@ class FileLevelManifest(ItemizedManifest):
         return self._digest_info == other._digest_info
 
 
-@dataclasses.dataclass(frozen=True)
+@dataclasses.dataclass
 class ShardedFileManifestItem(ManifestItem):
     """A manifest item that records a file shard together with its digest."""
 
@@ -130,6 +141,28 @@ class ShardedFileManifestItem(ManifestItem):
     start: int
     end: int
     digest: hashing.Digest
+
+    def __init__(
+        self,
+        *,
+        path: pathlib.PurePath,
+        start: int,
+        end: int,
+        digest: hashing.Digest
+    ):
+        """Builds a manifest item pairing a file with its digest.
+
+        Args:
+            path: The path to the file, relative to the model root.
+            start: The start offset of the shard.
+            end: The end offset of the shard.
+            digest: The digest of the file.
+        """
+        # Note: we need to force a PosixPath to canonicalize the manifest
+        self.path = pathlib.PurePosixPath(path)
+        self.start = start
+        self.end = end
+        self.digest = digest
 
     @property
     def input_tuple(self) -> tuple[pathlib.PurePath, int, int]:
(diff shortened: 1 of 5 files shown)


susperius:
> CC @susperius: this is the first step in the manifest generation. I'm currently working on serializing this to in-toto and other formats so we can do the signature.

That's great, thank you!

Would it make sense to define the manifest as a proto and then use `betterproto`? It would allow us to easily serialize and de-serialize the manifests to/from disk. (if betterproto is not an option, simple protobuf to json is also enough)

model_signing/serialization/itemized.py
```diff
@@ -0,0 +1,228 @@
...
+    Traverses the model directory and creates digests for every file found,
+    sharding the file in equal shards and computing the digests in parallel.
+
+    Since the manifest lists each item individually, this will also enable
+    support for incremental updates (to be added later).
+    """
+
+    def __init__(
+        self,
+        hasher_factory: Callable[
```
susperius, line 127: nit: Rename this to `sharded_hasher_factory` and the above one to `file_hasher_factory` to reduce confusion.

model_signing/serialization/itemized.py
```diff
@@ -0,0 +1,228 @@
...
+    possibly in parallel.
+
+    Since the manifest lists each item individually, this will also enable
+    support for incremental updates (to be added later).
+    """
+
+    def __init__(
+        self,
+        hasher_factory: Callable[[pathlib.Path], file.FileHasher],
+        max_workers: int | None = None,
```
susperius, line 41: How do you feel about using `Optional[int]`?
It avoids the duplicated `None` :)

model_signing/serialization/itemized.py
```diff
@@ -0,0 +1,228 @@
...
+        Args:
+            model_path: The path to the model.
+            path: Path to the file in the model, that is currently transformed
+              to a manifest item.
+
+        Returns:
+            The itemized manifest.
+        """
+        # Note: we need to force a PosixPath to canonicalize the manifest
+        relative_path = pathlib.PurePosixPath(path.relative_to(model_path))
```
susperius, line 100: It might make sense to move this into the manifest class itself. So, a user cannot mishandle it.  Perhaps adding a custom init method to the dataclasses that take `path` and `model_path` as arguments and then makes the call to `PurePosixPath`.

mihaimaruseac: Oh, this is good idea, thank you. I'll try it once I get the unit tests working (need to convert the paths to Posix there too, after I alter the model)

model_signing/serialization/itemized.py
mihaimaruseac: The internal coding style now recommends the `int | None` version as `Optional[int]` could be read as both an argument that is optional and an argument that can be None.

model_signing/serialization/itemized.py
mihaimaruseac: Actually, to do so we would have to unfreeze the dataclass. Let me think a little bit more if that's desired.

mihaimaruseac:
I'll experiment with `betterproto` among the designs for the next stage. Thnak you for the suggestion

model_signing/manifest/manifest.py
```diff
@@ -37,3 +73,113 @@ class DigestManifest(Manifest):
...
+        digest: hashing.Digest
+    ):
+        """Builds a manifest item pairing a file with its digest.
+
+        Args:
+            path: The path to the file, relative to the model root.
+            start: The start offset of the shard.
+            end: The end offset of the shard.
+            digest: The digest of the file.
+        """
```
spencerschrock, line 160: may be a minor copy/paste here in terms of what the digest applies to.

> Builds a manifest item pairing a **file** with its digest.

> digest: The digest of the **file**.

I believe these should both be "file shard"

mihaimaruseac: Nice catch, fixed!

model_signing/manifest/manifest.py
```diff
@@ -14,14 +14,51 @@
...
+manifest then gets signed and the signed manifest (or just signature) is stored
+in a signature file determined from the model path and type. The signature is
+loosely tied with the manifest (e.g., it could be a field in the manifest or it
+could be a signature over a single digest).
+
+When testing the integrity of a model, we start with the model path and model
+type. From this, we extract the path to the signed manifest. Next step is to
+verify the authenticity of the signature, after which we extract an in-memory
+representation of the _trusted_ manifest. To verify the model in its integrity,
+we run another `serialization.Serializer` instance to compute the _untrusted_
```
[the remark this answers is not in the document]
mihaimaruseac: Changed to expected vs actual.

model_signing/manifest/manifest.py
```diff
@@ -14,14 +14,51 @@
 
 """Machinery for representing a serialized representation of an ML model.
 
-Currently, we only support a manifest that wraps around a digest. But, to
-support incremental updates and partial signature verification, we need a
-manifest that lists files and their digests. That will come in a future change,
-soon.
+When saving a model, or signing an existing one, we build a manifest from the
```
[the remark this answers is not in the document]
mihaimaruseac: Yeah, the plan is to move parts of this to the README or somewhere where we document how the library works. For now, it's mostly my notes on how we're thinking this to work.

model_signing/manifest/manifest.py
```diff
@@ -14,14 +14,51 @@
...
+model (either by using a `serialization.Serializer` or building it from helper
+methods that use precomputed values to reduce amount of I/O involved). This
+manifest then gets signed and the signed manifest (or just signature) is stored
+in a signature file determined from the model path and type. The signature is
```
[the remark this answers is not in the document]
mihaimaruseac: Reworded to make it slightly clearer. Thnaks for reviewing!

model_signing/manifest/manifest.py
```diff
@@ -14,14 +14,48 @@
...
+When testing the integrity of a model, we start with the model path and model
+type. From this, we extract the path to the signed manifest. Next step is to
+verify the authenticity of the signature, after which we extract an in-memory
+representation of the _expected_ manifest. To verify the integrity of the model,
+we run another `serialization.Serializer` instance to compute the _actual_
+manifest of the model as we're seeing it. If these two manifests agree then the
+model integrity is maintained.
+
+In the simplest case, we are working with `DigestManifest` objects. Here, the
+two manifests are conforming iff the digests are the same. A more complex case
```
smeiklej, line 31:
```suggestion
two manifests agree if and only if the digests are the same. A more complex case
```

(it looks like you otherwise changed 'conform' to 'agree')

mihaimaruseac: Thanks. Right, I missed this one

susperius (approved):

model_signing/manifest/manifest.py
```diff
@@ -37,3 +71,110 @@ class DigestManifest(Manifest):
...
+        start: int,
+        end: int,
+        digest: hashing.Digest
+    ):
+        """Builds a manifest item pairing a file shard with its digest.
+
+        Args:
+            path: The path to the file, relative to the model root.
+            start: The start offset of the shard.
+            end: The end offset of the shard.
```
laurentsimon, line 156: nit: maybe say whether the byte at position end is included or not

laurentsimon (approved):

model_signing/manifest/manifest.py
mihaimaruseac: Done :)

laurentsimon (approved):

Issue #231, opened 2024-07-09
Lint over the entire repository


mihaimaruseac:
Currently, some of the linting only happens for the model signing part of the repository, only for the new files for the API (e.g., the pytype linting). We should cover all of the codebase.

spencerschrock:
Briefly used `ruff` to address an import ordering comment left during code review. It seems nice and configurable, and the options that match our current import ordering would be:

```toml
[lint.isort]
force-single-line = true
force-sort-within-sections = true
single-line-exclusions = ["collections.abc", "typing"]
```

PR #232, opened 2024-07-09
Replace flake8 with pylint to be more attuned to Google's style

mihaimaruseac:
#### Summary
Use pylint instead of flake8 for linting, to match internal Google style. This is just a temporary step before we can add `pyproject.toml` and actually run `ruff`. See also #231 

Doing this because the pylint annotation in #230 are ignored by flake8.

#### Release Note
NONE

#### Documentation
NONE


.github/workflows/lint.yml
```diff
@@ -8,20 +8,16 @@ on:
...
     runs-on: ubuntu-latest
-    name: Lint
+    name: Lint whitespace
     steps:
       - name: Check out source repository
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
       - name: Set up Python environment
         uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
         with:
           python-version: "3.11"
```
spencerschrock, line 20: we should be able to remove actions-setup/python after removing flake8?

.github/workflows/lint.yml
```diff
@@ -71,4 +68,25 @@ jobs:
...
+          python-version: "3.11"
+      - name: run pylint
+        run: |
+          python -m venv venv
+          .github/workflows/scripts/venv_activate.sh
+          pip install -r model_signing/install/requirements_test_Linux.txt
+          pip install -r model_signing/install/requirements_dev_Linux.txt
+          # TODO: https://github.com/sigstore/model-transparency/issues/231 - Support all repo
+          # We should actually migrate to ruff, but that's configured via pyproject.toml which we use when we release the wheel
+          pylint -dC0114,C0115,C0116,R0903,R0904,R0913,R0914,R1721,R1737,W0107,W0223,W0231,W0511,W0621 model_signing/{hashing,manifest,serialization}
```
spencerschrock, line 92: let's prefer the long form for clarity
```shell
pylint --disable=...
```

spencerschrock (approved):

.github/workflows/lint.yml
mihaimaruseac: Good idea. Removed

.github/workflows/lint.yml
mihaimaruseac: Oh, that's clearer

spencerschrock (approved):

PR #242, opened 2024-07-19
Reduce fixture scope to class instead of session.

mihaimaruseac:
#### Summary
Session scope means the fixtures are created at the start of `pytest` run and finalized when the testing ends. This means their lifetime is much longer than what we need.

Class scope means that fixtures are created when the test class is first initialized and finalized when all the tests under that class have finished. This is more appropriate to what we want to use these fixtures for.

#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/model_signing/hashing/file_test.py
+++ b/model_signing/hashing/file_test.py
@@ -27,21 +27,21 @@ _SHARD_SIZE = len(_HEADER)
 _UNUSED_PATH = pathlib.Path("unused")
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="class")
 def sample_file(tmp_path_factory):
     file_path = tmp_path_factory.mktemp("dir") / "text.txt"
     file_path.write_text(_FULL_CONTENT)
     return file_path
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="class")
 def sample_file_content_only(tmp_path_factory):
     file_path = tmp_path_factory.mktemp("dir") / "text.txt"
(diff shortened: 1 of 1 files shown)


model_signing/hashing/file_test.py
```diff
@@ -27,21 +27,21 @@
...
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="class")
 def sample_file_content_only(tmp_path_factory):
     file_path = tmp_path_factory.mktemp("dir") / "text.txt"
     file_path.write_text(_CONTENT)
     return file_path
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="class")
```
spencerschrock, line 44: For these expected digests ones, does it matter if the expected digest lasts longer? It's not like we expect the value to change based on which test class calls it?

mihaimaruseac: They're tied with these files, it does not make sense to have them around once these files are no longer present.

These fixtures should only exist while testing the hashing layer, once we move to the serialization layer we have a different set of fixtures for the models. There, because we alter the models, we use the default scope, where each fixture is created before each test and destroyed at the end of each one. We don't mutate these here, so test scope is too much. Next possible one is class scope.

spencerschrock (approved):

model_signing/hashing/file_test.py
spencerschrock: Ah, I had wrongly assumed this fixture was used elsewhere

PR #250, opened 2024-07-23
fix --update_goldens for serialization module

spencerschrock:
#### Summary
Writing the sharded serializer golden files involves sorting `_Shard` items, so they must be orderable. Previous attempts to update goldens would result in a `TypeError`:

#### Release Note

NONE

#### Documentation

NONE


mihaimaruseac (approved):
Thank you!

Internally this doesn't fail, and I only included the fix in #245 instead of the good PR.

Thanks for catching!

spencerschrock:
> Internally this doesn't fail

It does actually. The `_Shard` dataclass was added after I asked you to regenerate the goldens during review

mihaimaruseac:
Oh, I only ran the tests at the top of the chain :| That's on me

Issue #196, opened 2024-06-04
Handle symlinks and excluded files in serialization


mihaimaruseac:
**Description**
Following #190: during migration I did not yet add support for symlinks and for excluding files. This will be handled later when we find a need for it during end-to-end testing.

spencerschrock:
> I did not yet add support for symlinks

To clarify, I believe the code already supports symlinks as written (see #251), the TODO refers to preventing symlinks.
https://github.com/sigstore/model-transparency/blob/90e3ed354016309cb6ceca1b47bc4b8b23deb97e/model_signing/serialization/serialize_by_file_shard.py#L126-L129

This restriction was added in #24, and the only reason I can tell is from [this block](https://github.com/sigstore/model-transparency/blob/90e3ed354016309cb6ceca1b47bc4b8b23deb97e/model_signing/serialize.py#L151-L157), which reads as a TODO to potentially allow symlinks:

> To avoid bugs where we read the link rather than its target, 
we don't allow symlinks for now.
NOTE: It seems that Python's read() *always* follows symlinks,
so it may be safe to allow them. (readlink() is the function
to read the link metadata).

It would be simple to add a toggle and check for them, but is it needed? Would a signer/verifier care about the presence of a symlink?

mihaimaruseac:
There might be an issue when the model directory has a symlink inside it that points to a file outside. If we blindly read the symlink, we won't be able to verify the model after it gets uploaded, as that file won't be sent over.

If we add a toggle to error when a symlink is found, we can detect this scenario earlier.

I don't know if there are other cases where we'd care. If the model contains a symlink to a file within itself, nothing should change

spencerschrock:
> we won't be able to verify the model after it gets uploaded, as that file won't be sent over.

I assume this is heavily dependent on the upload process for a given model hub? My symlink knowledge is superficial at best, so I did a naive test  with `scp` from macOS -> Linux which seemed to have copied the contents.:

```shell
mkdir /tmp/foo
echo "HELLO" > /tmp/foo/bar.txt
mkdir /tmp/symlink-test
ln -s /tmp/foo/bar.txt /tmp/symlink-test/baz.txt
scp -r /tmp/symlink-test <host>:~
```

And then on the host, I see a normal file with the expected contents:
```shell
$ cat symlink-test/baz.txt 
HELLO
$ ls -laht symlink-test/baz.txt 
[...]

PR #257, opened 2024-07-25
combine model serialization logic for files and directories

spencerschrock:
#### Summary

Identical logic was being performed whether model_path pointed to a file or a directory due to two pathlib.Path.glob behaviors:

1. The path must be a directory for the generator to yield items
2. The glob does not include the provided directory itself.

By iterating over the provided model_path first, and then continuing to the glob, all of the serialization logic can be done in the loop. This will help future changes which will add more serialization logic in the form of ignore paths (#196).

#### Release Note

NONE

#### Documentation

NONE

### DIFF
--- a/model_signing/serialization/serialize_by_file.py
+++ b/model_signing/serialization/serialize_by_file.py
@@ -130,9 +130,9 @@ class FilesSerializer(serialization.Serializer):
         # Python3.12 is the minimum supported version, the glob can be replaced
         # with `pathlib.Path.walk` for a clearer interface, and some speed
         # improvement.
-        def root() -> Iterable[pathlib.Path]:
-            yield model_path
-        for path in itertools.chain(root(), model_path.glob("**/*")):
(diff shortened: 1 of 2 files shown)


mihaimaruseac (dismissed):
I like this!

One question to avoid the inner function, but if that doesn't seem clearer let's go with this

spencerschrock:
> One question to avoid the inner function, but if that doesn't seem clearer let's go with this

Do you have a suggestion? [itertools.chain](https://docs.python.org/3/library/itertools.html#itertools.chain) expects an iterator, and this was the syntax I found for a single path. Happy to switch to something better if there is something

mihaimaruseac:
Oh, sorry, somehow my inline comment didn't get saved. I probably forgot to click some button :(.

I was testing with `itertools.repeat(model_path, 1)`.

mihaimaruseac (approved):
Thank you and sorry again for not saving the other comment :(

PR #258, opened 2024-07-25
reduce pylint max line length to 80 from default of 100

spencerschrock:
#### Summary

This matches the limit previously used with flake8 before it was replaced in #232.

#### Release Note

NONE

#### Documentation

NONE


mihaimaruseac (approved):
Thank you. I didn't know the default was 100.

PR #260, opened 2024-07-26
WIP adopting manifests and signing classes

susperius:
WIP do not merge

This is my first draft of adopting the new signing and manifest primitives for actual signature creation and serialization to disk. For now the idea is to use `BytesSigner` to sign the contents of the manifest payload and provide verification material. That can then be stored on disk together with the signed payload as a tuple of (SigstoreBundle, Manifest).

@mihaimaruseac PLMK what you think about this approach


model_signing/signature/signature.py
```diff
@@ -0,0 +1,83 @@
...
+        Returns:
+            bundle_pb.VerificationMaterial: Sigstore verification material 
+        """
+
+
+class BytesSigner(abc.ABC):
+    """BytesSigner is the abstract base class for signing bytes."""
+
+    @abc.abstractmethod
+    def sign(self, data: bytes) -> bytes:
```
mihaimaruseac, line 47: This is the same as [`signing.Signer`](https://github.com/sigstore/model-transparency/blob/2198c9d471be3058e2d2fced7e6ddb5d3c116b5f/model_signing/signing/signing.py#L102). I'm curious why we'd duplicate the class hierarchy instead of using it?

model_signing/signature/signature.py
```diff
@@ -0,0 +1,83 @@
...
+        """
+        Returns the verification material needed to verify signatures
+        generated by this signer.
+
+        Returns:
+            VerificationMaterial: Material used to verify signatures.
+        """
+
+
+class BytesVerifier(metaclass=abc.ABCMeta):
```
mihaimaruseac, line 69: Same here, w.r.t `signing.Verifier`

model_signing/signature/signature.py
```diff
@@ -0,0 +1,83 @@
...
+
+from in_toto_attestation.v1 import statement
+from sigstore_protobuf_specs.dev.sigstore.bundle import v1 as bundle_pb
+
+
+class VerificationMaterial(metaclass=abc.ABCMeta):
+    """Generic verification material"""
+
+
+class SigstoreVerificationMaterial(VerificationMaterial):
```
mihaimaruseac, line 25: Can this be just a subclass of `signing.Signature, which will just wrap the bundle?

model_signing/signing/manifest.py
```diff
@@ -0,0 +1,82 @@
...
+            r.identifier: r.digest for r in resource_descriptors
+        }
+
+    @classmethod
+    @override
+    def from_manifest(cls, manifest: Manifest) -> Self:
+        descriptors = []
+        for d in manifest.resource_descriptors:
+            descriptors.append(d)
+        return cls(descriptors)
```
mihaimaruseac, line 42: I'm unsure about this. We're just building a dictionary out of items fed from another dictionary. And we also lose the ordering guarantee.

Why not make the payload directly be in-toto here?

mihaimaruseac (commented):
Sorry, just saw this.

I was working on this area too, as said via chat. I'll push my PR shortly, once I clean it. But I think we can go directly to in-toto now, we don't need to replicate the manifest once more

mihaimaruseac:
If you want to see the entire work to convert manifests to in-toto, #267 (I split it into a series of PRs with various in-toto payloads to ease reviews - every PR should just look at the last commit).

susperius:
> Sorry, just saw this.
> 
> I was working on this area too, as said via chat. I'll push my PR shortly, once I clean it. But I think we can go directly to in-toto now, we don't need to replicate the manifest once more

ok, I'll leave it at that and stop trying to contribute for now.

mihaimaruseac:
That wasn't my intention :(

Issue #248, opened 2024-07-23
Support for converting manifests to in-toto statements


susperius:
**Description**

Tracking bug for manifest to in-toto statement conversion.

In order to sign and store model signatures in a sigstore bundle the manifest must be converted to an in-toto statement.

mihaimaruseac:
Included in my draft work, will send a draft PR

laurentsimon:
I've had some thoughts about the intoto format. In https://github.com/sigstore/model-transparency/issues/111, we expressed concern over listing each file as intoto subjects, because tooling like cosign can be used to (wrongly) verify only a subset of subjects. Here's an idea that may be able to address this concern:

1. We list the (file, (custom?) hash) inside the predicate, not the intoto subject
2. We output a _single_ subject in the intoto subject list by serializing the list of (file, hash), like we [currently do in our existing PoC](https://github.com/sigstore/model-transparency/blob/main/model_signing/serialization/serialize_by_file.py#L71). We need not serialize json (which requires canonicalization), but a simple format like `base64(path).hash-value` as shown [here](https://github.com/sigstore/model-transparency/blob/main/model_signing/serialization/serialize_by_file.py#L71). Other metadata for each file need not be serialized but are still protected by the manifest signature. Serialization schemes may vary, so we'll need a way to convey this as well. There are 2 ways to achieve this:
  a. In subject, by changing the hash scheme `scheme-sha256`.  This is a bit harder to support with sigstore-python](https://github.com/sigstore/sigstore-python/issues/1018), unless we build the protobuf ourselves.
  b. We add the scheme type under the predicate. I lean towards this solution.

So overall it would look like the following:

```json
{
  "_type": "https://in-toto.io/Statement/v0.1",
  "predicateType": "https://something/model-signing/v1",
  "subject": [
    {
      "digest": {
        "sha256": "blabla"
      }
    },
  ],
  "predicate": {
     "serialization": "some-name/v1",
     "files": {
           {
              "path": "to/file1",
              "digest": {
                 "sha256": "bla",
              }
           },
          {
              "path": "to/file2",
              "digest": {
                 "custom": "bla",
              }
           },
      },
   }
}
```

susperius:
I agree with you. It only feels like "abusing" the format which is mostly because sigstore has hard coded the acceptable hashes in their code base.

Perhaps it would be the best to go down the route of our own format to avoid additional headache. 
[...]

Issue #224, opened 2024-06-27
Manifest file format


susperius:
I wanted to reignite the discussion around a manifest file format. So, I tried to define two potential solutions as shown below. 
The first part shows a simple solution to support a Merkle tree root hash and the per file hash approaches in an either / or fashion.
Here, we could also extend the `root_hash` field to a custom type if not all necessary information can be captured in the metadata section. 
The second part shows a solution that would support a mixed approach. I'm not sure it is necessary but just wanted to highlight it.

In both cases `metadata` is a simple `dict[str, str]` type that allows users to add arbitrary metadata to the Manifest. This could also be extended to support a strongly typed key, value solution if necessary.

The Manifest could be packaged in a DSSE envelope to follow their standard procedure for signing or (if sigstore is going to allow arbitrary data) we add a signature field to the Manifest file. I'd prefer going with the DSSE envelope since it's already well defined for this use case.

WDYT?

```proto
// Manifest supporting either digests or a root hash
message Manifest {
    map<str, str> metadata = 1;
    oneof model {
        ModelDigests model_digests = 2;
        bytes root_hash = 3;
    }
}

message ModelDigests {
    repeated Digest digests =1;
}

message Digest {
    string method = 1;
    string path = 2;
    bytes hash = 3;
}
// ================================================
// Manifest supporting mixed data
message Manifest {
    map<str, str> metadata = 1;
    ModelInformation model_information = 2;
}

message ModelInformation {
    oneof data {
        Digest digest = 1;
        RootHash root_hash = 2;
    }
}

message RootHash {
    string archive_information = 1;
    bytes hash = 2;
}

message Digest {
    string method = 1;
    string path = 2;
    bytes hash = 3;
}
```

susperius:
I'm going to prepare a manifest PR. That's based upon [PR193](https://github.com/sigstore/model-transparency/pull/193) and adds a last modified field to the `Digest` message.

mihaimaruseac:
I'm currently working on one, to handle internal BCID needs.

PR #263, opened 2024-07-27
Create in-toto signing payload for single digests.

mihaimaruseac:
#### Summary
This converts `DigestManifest`s objects to an in-toto format where the model is identified by its complete digest (as result of the serialization).

For Sigstore signing, this payload can be signed via `sign_intoto`, producing a Sigstore `Bundle` as the signature.

CC @susperius for converting manifest to in-toto. This should cover #111, #224, and #248 (first part of the machinery). CC @laurentsimon to make sure I did not mishandle in-toto.

**Note**: Looking at the goldens, I see that some of the hashes are not properly converted to the one we use in serialization (should always be `file-sha256` in the current testing scenario). I was planning to debug that before pushing this work, but given #260 I'll send it now to reduce work duplication and will debug the issue next week.

**Note**: Disabling C0103 and E1101 in the pylint configuration. The first one is about the name of the class constant, it wants it to be all uppercase, but that goes against established style for class/instance variables. Worse, the second one fails to detect that protobuf generated code has certain attributes/names. This is because [pylint does not use .pyi files, which is what protobuf generates](https://github.com/pylint-dev/pylint/issues/6281), but we can disable it given this missing attribute would be detected by pytype, which works correctly. Maybe when switching linter we'd get rid of this issue.

#### Release Note
NONE
#### Documentation
NONE

### DIFF
--- a/model_signing/signing/in_toto.py
+++ b/model_signing/signing/in_toto.py
@@ -1,0 +1,125 @@
+# Copyright 2024 The Sigstore Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Signing payloads for models as in-toto statements.
+
+To generate the signing payload we convert model manifests to in-toto formats,
+as described by https://github.com/in-toto/attestation/tree/main/spec/v1. The
+envelope format is DSSE, see https://github.com/secure-systems-lab/dsse.
+"""
+
+from typing import Final, Self
+
+from in_toto_attestation.v1 import statement
+from typing_extensions import override
+
+from model_signing.manifest import manifest as manifest_module
+from model_signing.signing import signing
+
+
+class IntotoPayload(signing.SigningPayload):
+    """A generic payload in in-toto format.
+
+    This class is abstract for now as we will support multiple payload formats
+    below.
+
+    Each subclass defines a constant for the predicate type class attribute
+    defined below.
+    """
+
+    predicate_type: Final[str]
+
+
+class SingleDigestIntotoPayload(IntotoPayload):
+    """In-toto payload where the model is serialized to just one digest.
+
+    In this case, we encode the model as the only subject of the statement. We
+    don't set the name field, and use the digest as the one resulting from the
+    model serialization.
+
+    However, since we use custom hashing algorithms, but these are not supported
+    by existing tools, we claim that the digest algorithm is sha-256 and include
+    the real digest in the predicate.
+
+    Example:
+    ```json
+    {
+      "_type": "https://in-toto.io/Statement/v1",
+      "subject": [
+        {
+          "digest": {
+            "sha256": "3aab065c...."
+          }
+        }
+      ],
+      "predicateType": "https://model_signing/Digest/v0.1",
+      "predicate": {
+        "actual_hash_algorithm": "file-sha256"
+      }
+    }
+    ```
+
+    If the predicate is missing (or does not set "actual_hash_algorithm"), it
+    should be assumed that the digest is actually computed via the algorithm
+    present in the resource descriptor (i.e., sha256).
+
+    See also https://github.com/sigstore/sigstore-python/issues/1018.
+    """
+
+    predicate_type: Final[str] = "https://model_signing/Digest/v0.1"
+
+    def __init__(self, *, digest_hex: str, digest_algorithm: str):
+        """Builds an instance of this in-toto payload.
+
+        Don't call this directly in production. Use `from_manifest()` instead.
+
+        Args:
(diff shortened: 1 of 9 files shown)


model_signing/signing/in_toto.py
```diff
@@ -0,0 +1,125 @@
...
+            digest_hex: the hexadecimal, human readable, digest of the subject.
+            digest_algorithm: the algorithm used to compute the digest.
+        """
+        digest = {"sha256": digest_hex}
+        descriptor = statement.ResourceDescriptor(digest=digest).pb
+
+        self.statement = statement.Statement(
+            subjects=[descriptor],
+            predicate_type=self.predicate_type,
+            predicate={"actual_hash_algorithm": digest_algorithm},
```
susperius, line 97: Perhaps consider adding this as annotation to be aligned with the later PR.

model_signing/signing/in_toto.py
```diff
@@ -0,0 +1,125 @@
...
+    If the predicate is missing (or does not set "actual_hash_algorithm"), it
+    should be assumed that the digest is actually computed via the algorithm
+    present in the resource descriptor (i.e., sha256).
+
+    See also https://github.com/sigstore/sigstore-python/issues/1018.
+    """
+
+    predicate_type: Final[str] = "https://model_signing/Digest/v0.1"
+
+    def __init__(self, *, digest_hex: str, digest_algorithm: str):
```
susperius, line 82: Consider adding a metadata dict that can be used as predicate with a default {'unused': True} or something like that. It could be used to provide model information down the road.

model_signing/signing/in_toto.py
mihaimaruseac: We need a predicate. If I move this to an annotation then I'll have to add a useless predicate

model_signing/signing/in_toto.py
mihaimaruseac: Let's add that when needed

model_signing/signing/in_toto.py
```diff
@@ -0,0 +1,125 @@
...
+      "subject": [
+        {
+          "digest": {
+            "sha256": "3aab065c...."
+          }
+        }
+      ],
+      "predicateType": "https://model_signing/Digest/v0.1",
+      "predicate": {
+        "actual_hash_algorithm": "file-sha256"
```
laurentsimon, line 68: Note: We could also name this field "serialization_xxx" instead of `hash_xxx`. Might be more consistent with the codebase (we have distinct hash and serialization code), but more confusing since intoto does not this concept. On the other hand, I recall Hayden saying he'd like to see cosign and other tools have a dedicated serialization layer, iirc

Note 2: In SLSA we typically use camel case for all field names.

Those are just fyis, feel free to ignore

model_signing/signing/in_toto.py
```diff
@@ -0,0 +1,125 @@
...
+
+    predicate_type: Final[str] = "https://model_signing/Digest/v0.1"
+
+    def __init__(self, *, digest_hex: str, digest_algorithm: str):
+        """Builds an instance of this in-toto payload.
+
+        Don't call this directly in production. Use `from_manifest()` instead.
+
+        Args:
+            digest_hex: the hexadecimal, human readable, digest of the subject.
```
laurentsimon, line 88: nit: maybe say that `0x` is _not_ prepended to the string

model_signing/signing/testdata/in_toto/TestSingleDigestIntotoPayload/deep_model_folder
```diff
@@ -0,0 +1,14 @@
+{
```
laurentsimon, line 1: nit: maybe add the extension `.intoto` to the file, to give a hint to readers of what the file contains

laurentsimon (dismissed):

model_signing/signing/testdata/in_toto/TestSingleDigestIntotoPayload/deep_model_folder
mihaimaruseac: Let me do this after merging all of these, to do it across the board (rebase would be harder otherwise, with a lot of conflicts due to file moves).

model_signing/signing/in_toto.py
mihaimaruseac: Oh, we never do, we're consistent with the rest of the codebase.

model_signing/signing/in_toto.py
mihaimaruseac: So, `serializationAlgorithm` instead of `actual_hash_algorithm`? I'll do that after the set lands, otherwise due to having to regenerate goldens there will be a ton of conflicts.

laurentsimon (approved):

model_signing/signing/testdata/in_toto/TestSingleDigestIntotoPayload/deep_model_folder
spencerschrock: FYI: If you decide to rename, this would go against the [naming convention](https://github.com/in-toto/attestation/blob/main/spec/v1/envelope.md#file-naming-convention). So maybe `.intoto.json` ?

> If stored in a dedicated file by itself, and not as part of a [Bundle](https://github.com/in-toto/attestation/blob/main/spec/v1/bundle.md), an Envelope SHOULD use the suffix .json.

mihaimaruseac: Oh, that's correct. Yes, it should be `.intoto.json`

PR #264, opened 2024-07-27
Add in-toto format for items with hash of hashes

mihaimaruseac:
#### Summary

This converts model serialization manifests that record every model file hash into an in-toto payload that can then be passed to Sigstore's `sign_intoto` for signing to generate a Sigstore `Bundle` (if using Sigstore).

To identify the models, we compute a hash of all hashes of the files and use that as the subject. The individual file hashes are used as the payload and we would have the verifier check them as part of the verification process.

**Note**: This builds on #263. I decided to split every feature into its own PR to make it easier to review what changes (should be only the last commit) and to be able to merge partial work and continue from there.

#### Release Note
NONE
#### Documentation
NONE

### DIFF
--- a/model_signing/signing/in_toto.py
+++ b/model_signing/signing/in_toto.py
@@ -24,6 +24,7 @@ from typing import Final, Self
 from in_toto_attestation.v1 import statement
 from typing_extensions import override
 
+from model_signing.hashing import memory
 from model_signing.manifest import manifest as manifest_module
 from model_signing.signing import signing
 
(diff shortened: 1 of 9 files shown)


model_signing/signing/in_toto.py
```diff
@@ -0,0 +1,256 @@
...
+        })
+
+    digest = {"sha256": hasher.compute().digest_hex}
+    descriptor = statement.ResourceDescriptor(digest=digest).pb
+
+    return statement.Statement(
+        subjects=[descriptor],
+        predicate_type=predicate_type,
+        predicate={predicate_top_level_name: subjects},
+    )
```
susperius, line 160: Shouldn't the subject be a list of the file hashes?

It feels a bit like misusing the Statement format when the digests are put into the predicate.

mihaimaruseac: Doing that in #266. But in #111 there was a discussion that the semantics of verification on subjects is not working properly for models, so we put as subject a single digest for the entire model and then the actual hashes in payload to reconstruct as needed.

TomHennen:
I think my biggest concern with this proposal (and #265, #266, #267) is that it doesn't allow for any other predicates to be made for models in the future.  It can only sign things, but cannot convey any additional information with that signature (things like what types of data this model was trained on, etc...).

It sounds like the primary reason for this approach is that Sigstore cannot support other hash types in Rekor.  To me that seems like the more important problem to solve (but I don't understand the technical problems with doing so).

mihaimaruseac:
Right, these are just for signing, though we can extend the payload to contain more information, I think

model_signing/signing/in_toto.py
susperius: I don't really see the verification argument. It applies to every format you roll with, if there is another implementation there is always the risk of doing it wrong/different.

mihaimaruseac: I just implemented all manifest ideas discussed in #111 so we can experiment with them and see which one is more ergonomic, better suited for our needs, less prone to failures

mihaimaruseac: Also, this specific PR is what's being discussed in https://github.com/sigstore/model-transparency/issues/248#issuecomment-2248298175

laurentsimon (approved):

model_signing/signing/in_toto.py
```diff
@@ -123,3 +124,133 @@ def from_manifest(cls, manifest: manifest_module.Manifest) -> Self:
...
+    ```
+
+    A missing predicate, or a predicate for which an entry does not have valid
+    name, digest, or algorithm should be considered invalid and fail integrity
+    verification.
+
+    See also https://github.com/sigstore/sigstore-python/issues/1018.
+    """
+
+    predicate_type: Final[str] = "https://model_signing/DigestOfDigests/v0.1"
```
spencerschrock, line 219: I was curious if it mattered that this wasn't resolvable. [The answer is apparently no](https://github.com/in-toto/attestation/blob/main/spec/v1/field_types.md#typeuri), but should we namespace this any more to avoid collisions?


> SHOULD resolve to a human-readable description, but MAY be unresolvable. SHOULD include a version number to allow for revisions.
> 
> TypeURIs are not registered. The natural namespacing of URIs is sufficient to prevent collisions.

spencerschrock (approved):

model_signing/signing/in_toto.py
mihaimaruseac: Likely yes. Right now all these constants are just for testing. When we standardize, we will pick sensible normal values for them.

Thank you for flagging and the review!

PR #265, opened 2024-07-27
Add in-toto format as hash of shard hashes

mihaimaruseac:
#### Summary

This converts model serialization manifests that record every model file hash into an in-toto payload that can then be passed to Sigstore's `sign_intoto` for signing to generate a Sigstore `Bundle` (if using Sigstore).

To identify the models, we compute a hash of all hashes of the files and use that as the subject. The individual file hashes are used as the payload and we would have the verifier check them as part of the verification process.

**Note**: This is the equivalent of #264, but for file shard hashes instead of file hashes.

**Note**: This builds on #264. I decided to split every feature into its own PR to make it easier to review what changes (should be only the last commit) and to be able to merge partial work and continue from there.

#### Release Note
NONE
#### Documentation
NONE

### DIFF
--- a/model_signing/signing/in_toto.py
+++ b/model_signing/signing/in_toto.py
@@ -254,3 +254,101 @@ class DigestOfDigestsIntotoPayload(IntotoPayload):
             predicate_top_level_name="files",
         )
         return cls(statement)
+
+
+class DigestOfShardDigestsIntotoPayload(IntotoPayload):
+    """In-toto payload where the subject is a digest of digests of file shards.
+
+    This payload is supposed to be used for manifests where every file shard in
+    the model is matched with a digest. Because existing tooling only supports
+    established hashing algorithms, we record every such digest in the predicate
+    part and compute a hash for the subject by using sha256 on the concatenation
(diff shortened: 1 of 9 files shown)


model_signing/signing/in_toto.py
```diff
@@ -0,0 +1,354 @@
...
+    {
+      "_type": "https://in-toto.io/Statement/v1",
+      "subject": [
+        {
+          "digest": {
+            "sha256": "18b5a4..."
+          }
+        }
+      ],
+      "predicateType": "https://model_signing/DigestOfShardDigests/v0.1",
```
laurentsimon, line 280: I'm curious if the non-sharded version can be represented as a special case of this one, with the shard=0.
Not sure it's worth it to try to optimize it. Thinking more about it, I like that it's separate to avoid implementation problems :)

laurentsimon (dismissed):

model_signing/signing/in_toto.py
mihaimaruseac: Hmm, we can try that in the future.

spencerschrock (approved):

PR #267, opened 2024-07-27
Add in-toto format with hashes of shards as subjects

### LINKED ISSUE #111: Manifest file

### LINKED ISSUE #224: Manifest file format

### LINKED ISSUE #248: Support for converting manifests to in-toto statements

mihaimaruseac:
#### Summary

This converts model serialization manifests that record every model file shard hash into an in-toto payload that can then be passed to Sigstore's `sign_intoto` for signing to generate a Sigstore `Bundle` (if using Sigstore).

This time, we record every hash as part of the subject instead of in the payload. We require verifiers to be aware of this and acknowledge that verifiers that only check subject by subject (that is, they check if the hash of a passed in argument is in the list of subjects and don't check if all the hashes are present), can fail to fully detect if the model integrity is compromised by renaming one file in the model, interchanging two file names, deleting a file, or reordering two shards. The signing library will have additional checks for this, but verifying the signature with other tools might result in invalid results.

**Note**: I still had to pass some payload to the in-toto predicate due to https://github.com/in-toto/attestation/issues/374. Right now, it is a key-val pair that should be ignored, but there is also the possibility of registering only one predicate type for all model signing in-toto formats and then registering a subtype as part of the predicate, that we can control as much as we need.

**Note**: This is the equivalent of #264, but with digests added as subjects.

**Note**: This is the equivalent of #266, but for file shard hashes instead of file hashes.

**Note**: This builds on #266. I decided to split every feature into its own PR to make it easier to review what changes (should be only the last commit) and to be able to merge partial work and continue from there.

Finally, merging this will:
- Close #111
- Close #224 
- Close #248 
- Make #112 obsolete
- Probably require removing/re-doing #260 and #240.

#### Release Note
NONE
#### Documentation
NONE

### DIFF
--- a/model_signing/signing/in_toto.py
+++ b/model_signing/signing/in_toto.py
@@ -481,3 +481,108 @@ class DigestsIntotoPayload(IntotoPayload):
             manifest, predicate_type=cls.predicate_type
         )
         return cls(statement)
+
+
+class ShardDigestsIntotoPayload(IntotoPayload):
+    """In-toto payload where the subjects are the model shards themselves.
+
+    This payload is supposed to be used for manifests where every file shard in
+    the model is matched with a digest. Because existing tooling only supports
+    established hashing algorithms, we annotate every subject with the actual
+    hash algorithm used to compute the file digest, and use "sha256" as the
+    algorithm name in the digest itself.
+
+    Example:
+    ```json
+    {
+      "_type": "https://in-toto.io/Statement/v1",
+      "subject": [
+        {
+          "name": "d0/d1/d2/d3/d4/f0:0:16",
+          "digest": {
+            "sha256": "6efa14..."
+          },
+          "annotations": {
+            "actual_hash_algorithm": "file-sha256-1000000"
+          }
+        },
+        {
+          "name": "d0/d1/d2/d3/d4/f1:0:16",
+          "digest": {
+            "sha256": "a9bc14..."
+          },
+          "annotations": {
+            "actual_hash_algorithm": "file-sha256-1000000"
+          }
+        },
+        {
+          "name": "d0/d1/d2/d3/d4/f2:0:16",
+          "digest": {
+            "sha256": "5f597e..."
+          },
+          "annotations": {
+            "actual_hash_algorithm": "file-sha256-1000000"
+          }
+        },
+        {
+          "name": "d0/d1/d2/d3/d4/f3:0:16",
+          "digest": {
+            "sha256": "eaf677..."
+          },
+          "annotations": {
+            "actual_hash_algorithm": "file-sha256-1000000"
+          }
+        }
+      ],
+      "predicateType": "https://model_signing/ShardDigests/v0.1",
+      "predicate": {
+        "unused": "Unused, just passed due to API requirements"
+      }
+    }
+    ```
+
+    If the annotation for a subject is missing, or it does not contain
+    actual_hash_algorithm, it should be assumed that the digest is computed via
+    the algorithm listed in the digest dictionary (i.e., sha256).
+
+    See also https://github.com/sigstore/sigstore-python/issues/1018.
+    """
+
+    predicate_type: Final[str] = (
+        "https://model_signing/ShardDigests/v0.1"
+    )
+
+    def __init__(self, statement: statement.Statement):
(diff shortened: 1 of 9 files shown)


laurentsimon (dismissed):

spencerschrock (approved):

Note: In the PR description I'm assuming

> Note: This is the equivalent of https://github.com/sigstore/model-transparency/pull/265, but for file shard hashes instead of file hashes.

Should be

> Note: This is the equivalent of https://github.com/sigstore/model-transparency/pull/266, but for file shard hashes instead of file hashes.

mihaimaruseac:
Thank you for the review!

> Note: In the PR description I'm assuming
> 
> > Note: This is the equivalent of #265, but for file shard hashes instead of file hashes.
> 
> Should be
> 
> > Note: This is the equivalent of #266, but for file shard hashes instead of file hashes.

It should have been 2 notes, one describing the diff from #264 (subjects instead of payload) and one from #266. Fixed now.

PR #273, opened 2024-07-31
Replace pylint with ruff

spencerschrock:
#### Summary

Configure `ruff` with a barebones `pyproject.toml` to address the conversion mentioned in #232. I spent the time to do the switch now so I would stop messing up import ordering.

In terms of the rules selected, https://github.com/sigstore/model-transparency/commit/7f85551a332806989291ab144a0c2264c2a72f82 talks about the motivation for picking them, but I'm happy to adjust as needed, or include the reasons in the config file itself.
```toml
select = [
    "B", # flake8-bugbear
    "E", # pycodestyle
    ...
]
```

I wasn't sure where to place the `pyproject.toml` but we have options (including making it a `.ruff.toml` for now):

1. The repository root, which would allow us to configure `ruff` in one spot for both `model_signing` and `slsa_for_models`. The `tool.ruff` section can extend other `pyproject.toml`s, so if we had a `pyproject.toml` for each project they [could inherit `ruff` settings](https://docs.astral.sh/ruff/configuration/#config-file-discovery).
2. The `model_signing` directory (which is where it currently is in this PR).

For reviewers, I recommend going commit by commit, I tried to organize them in reviewable chunks.

@mihaimaruseac does this address the comments you left in #232?
> Doing this because the pylint annotation in https://github.com/sigstore/model-transparency/pull/230 are ignored by flake8.
 

#### Release Note

NONE

#### Documentation

NONE

### DIFF
--- a/model_signing/pyproject.toml
+++ b/model_signing/pyproject.toml
@@ -1,0 +1,12 @@
+[tool.ruff]
+line-length = 80
+target-version = "py311"
+
+[tool.ruff.lint]
+select = ["B", "E", "F", "I", "N", "PLC", "PLE", "SIM", "W"]
+ignore = ["B024"]
+
+[tool.ruff.lint.isort]
+force-single-line = true
+force-sort-within-sections = true
+single-line-exclusions = ["collections.abc", "typing"]
--- a/model_signing/serialization/serialize_by_file.py
+++ b/model_signing/serialization/serialize_by_file.py
@@ -98,10 +98,7 @@ def _ignored(path: pathlib.Path, ignore_paths: Iterable[pathlib.Path]) -> bool:
     Returns:
         Whether or not the provided path should be ignored.
     """
-    for ignore_path in ignore_paths:
-        if path.is_relative_to(ignore_path):
-            return True
-    return False
+    return any(path.is_relative_to(ignore_path) for ignore_path in ignore_paths)
 
 class FilesSerializer(serialization.Serializer):
     """Generic file serializer.
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -68,7 +68,7 @@ jobs:
           # TODO: https://github.com/sigstore/model-transparency/issues/231 - Support all repo
           pytype --keep-going model_signing/{hashing,manifest,serialization,signature,signing}
 
-  pylint-lint:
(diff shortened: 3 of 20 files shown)


model_signing/serialization/serialize_by_file.py
```diff
@@ -98,10 +98,7 @@ def _ignored(path: pathlib.Path, ignore_paths: Iterable[pathlib.Path]) -> bool:
     Returns:
         Whether or not the provided path should be ignored.
     """
-    for ignore_path in ignore_paths:
-        if path.is_relative_to(ignore_path):
-            return True
-    return False
+    return any(path.is_relative_to(ignore_path) for ignore_path in ignore_paths)
```
mihaimaruseac, line 101: Nice refactor!

model_signing/pyproject.toml
```diff
@@ -0,0 +1,12 @@
+[tool.ruff]
+line-length = 80
+target-version = "py311"
+
+[tool.ruff.lint]
+select = ["B", "E", "F", "I", "N", "PLC", "PLE", "SIM", "W"]
+ignore = ["B024"]
```
mihaimaruseac, line 7: We'll probably fix this for later, make sure all abstract classes have methods marked as abstract. But not for this PR

mihaimaruseac (approved):
Thank you!

Comments are not actionable for this PR, just remarks.

I think we should go with `project.toml` inside each directory, as for SLSA we might actually have multiple subpackages in the future. So doing it like here is :+1:

model_signing/serialization/serialize_by_file.py
spencerschrock: Identified and fix provided by one of the SIM rules :)

model_signing/pyproject.toml
spencerschrock: Sure. Specifically it took issue with the empty class here:
https://github.com/sigstore/model-transparency/blob/90cac6c4634300a6cbe54f4cdd011290dc1062dd/model_signing/manifest/manifest.py#L128-L140

mihaimaruseac: Oh, I assumed all of empty means trivially true, but it seems they want at least one method. Well, this will get fixed soon anyway

PR #278, opened 2024-08-06
fix issuer creation when provided an OIDC issuer

spencerschrock:
#### Summary

`sigstore.oidc.Issuer.production()` doesn't accept arguments, but the class initializer does. This should have been caught by a linter, but we have two upstream issues working against us.

pylint's `too-many-function-args` (E1121) would have caught this, but we switched to ruff in f8238f1ce41a1fdf08c82c0d2bed0fe898d63665 which doesn't support E1121 yet.
https://www.github.com/astral-sh/ruff/issues/970

pytype doesn't use library type information unless a stub is present in the typeshed repo (sigstore-python follows PEP 561 instead). https://www.github.com/google/pytype/issues/151

#### Release Note

NONE

#### Documentation

NONE

### DIFF
--- a/model_signing/signing/sigstore.py
+++ b/model_signing/signing/sigstore.py
@@ -118,7 +118,7 @@ class SigstoreSigner(signing.Signer):
         else:
           self._signing_context = sigstore_signer.SigningContext.production()
           if oidc_issuer is not None:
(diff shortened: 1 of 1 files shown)


mihaimaruseac (approved):
Thank you!

PR #251, opened 2024-07-23
prevent serialization of symlinks by default

spencerschrock:
#### Summary

Adds a serializer initializer argument (`allow_symlink`) to control whether symlinks raise a `ValueError` during serialization. 

For rationale, see https://github.com/sigstore/model-transparency/issues/196#issuecomment-2245977345

#### Release Note

NONE

#### Documentation

NONE

PR #281, opened 2024-08-12
enable flake8-pytest-style linter

spencerschrock:
#### Summary

Since we use `pytest`, we may as well enable the associated `flake8` linter. It caught [one best practice](https://docs.astral.sh/ruff/rules/pytest-raises-too-broad/) I didn't follow in #251. Additionally, it's one step closer to completing #231 as the linter now checks our `conftest.py`

With regard to `fixture-parentheses = false` configuration, both `@pytest.fixture` and `@pytest.fixture()` are valid, but the default settings expect parentheses and our code doesn't. We could also disable the rule entirely, but this will enforce consistency.

#### Release Note

NONE

#### Documentation

NONE


mihaimaruseac (approved):
Good catch! Thank you!

PR #279, opened 2024-08-07
Setup automated generation of model_signing documentation

spencerschrock:
#### Summary

Adds a `docs` Hatch environment, and an associated workflow to upload the documentation to GitHub Pages like [sigstore-python](https://sigstore.github.io/sigstore-python/sigstore.html).

Fixes #209

#### Release Note

NONE

#### Documentation

This change is the documentation :)

Issue #201, opened 2024-06-10
Add support to hash via GPUs


mihaimaruseac:
It should be possible to have a (set of) hashing engine(s) that could compute a digest by using GPU. These should be subclasses of `HashingEngine`.

Note: This follows from #140

Issue #175, opened 2024-05-13
v1 release


laurentsimon:
This is a proposal for v1 release.
Scope: support sigstore for path only (no in-memory path support). This requires to implement only a subset of https://github.com/sigstore/model-transparency/issues/172:
```python
# model.py
def sign(
      input: Path,
      sig: Path  = defaultSigPath(input),
      recompute_paths: []Path = None,
      ignored_paths: []Path = [".git"],
     ) -> None
    ...

def verify(
      input: Path,
      sig: Path  = defaultSigPath(input),
      verifier: Verifier = SigstoreVerifier(),
      ignored_paths: []Path = [".git"],
    ) -> bool

```

What needs to be done:
### Simple
- [ ] Make all API private except SigstoreSigner / SigstoreVerifier and model.verify / model.sign - see https://github.com/sigstore/model-transparency/issues/172 for APIs. 
- [x] https://github.com/sigstore/model-transparency/issues/174.
- [x] https://github.com/sigstore/model-transparency/issues/157
- [x] Add raw signer to API
- [x] https://github.com/sigstore/model-transparency/issues/179
- [x] https://github.com/sigstore/model-transparency/issues/196

### Medium
- [ ] https://github.com/sigstore/model-transparency/issues/160.
 
### More work
- [x] Decide the default parameters for sha256pv1 hash (shard size), or
- [x] https://github.com/sigstore/model-transparency/issues/13https://github.com/sigstore/model-transparency/issues/13. TBD: benchmarking
- [ ] Define the manifest format |https://github.com/sigstore/sigstore-python/issues/982

We can add support for more features https://github.com/sigstore/model-transparency/issues/172 without breaking backward compatibility

laurentsimon:
@mihaimaruseac what do you think of making these the prio for v1?

mihaimaruseac:
I think anything that comes for the API are priority. So #172 , #111 (manifest), #140 (hash engine), #160 (incremental rehashing) are a must. As part of this, we will also need to clean-up `BaseResult` (#174).

Looking back, that's exactly the list you mentioned above.

mihaimaruseac:
I think #157 might also be good to have in the release

mihaimaruseac:
Added all of them to a milestone too. But also added improvement issues there, so it's good to keep track of the important ones on this issue

laurentsimon:
[...]

PR #283, opened 2024-08-13
Prepare for wheel publishing (alpha patch version of 0.0.1)

### LINKED ISSUE #68: Re-factor unit tests
We should also consider resolving https://github.com/google/model-transparency/pull/64#discussion_r1378198940, which suggests reformatting some of the tests.

### LINKED ISSUE #231: Lint over the entire repository

mihaimaruseac:
#### Summary
Prepare the project so that we can publish a wheel with the library (as part of #175). Attempt to make sure we can still proceed with both building the library and the surrounding CLI.

We are going to use `hatch` for all development work as that makes it very simple:
- to test: `hatch test`
- to test with coverage: `hatch test -c`
- to test against all supported python versions: `hatch test --all` (`-c` still supported)
- to format the code: `hatch fmt`
- to check the formatting (in GHA, etc.): `hatch fmt --check`.
- to check typing: `hatch run type:check`. This is a custom environment and we can create others in the future (e.g., for documentation)
- to build the wheel: `hatch build`. The wheel can then be installed in a separate virtual env if needed for future testing.
- to run custom tests: `hatch shell` or `hatch run $cmd`.

Testing now happens in parallel, and in randomized fashion. Fast and removes the possibility of a test case influencing the behavior of another.

Another benefit of using `hatch` is that it allows us to build C/C++ extensions (via plugins), so when we need to get extra performance or drop to C/C++ (e.g., for #201) we can just do it.

---

In order to make this buildable as a wheel, I needed to restructure the code. So, the first commit moves files around and only changes 2 markdown links. The files are moved to get to a [`src/` layout](https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/#src-layout-vs-flat-layout). There are more benefits for a `src/` layout than a flat
layout, such as:

- documentation is easier to generate
- we won't accidentally test against code that is different than what users would see
- we can include only the source in the wheel, reducing PyPI bandwidth
- it makes it clearer which parts of the repo go into the wheel and which ones are ancillary. Incidentally, we might now want to move the SLSA parts to another repo and maybe rename this one? We can do this at a later time.

Since we would need to change `README.model_signing.md` anyway once the API is ready, I have not updated it besides just fixing the image pointer.

I have deleted old files from when this was a POC. Since we're working on releasing the library, we no longer need those. The code in these files is already implemented in the library. What these files used to test is now in the test files.

The alternative to not moving the files to the root of the folder would have been to duplicate `LICENSE`, etc. inside the `model-signing` directory, so that packaging tools would see it. But we still needed to move files around to get to the `src/` layout and to make documentation work correctly. So, it did not make sense to also duplicate `LICENSE`.

---

For the second commit of the PR I just focused on making sure that `hatch build` builds the wheel and the tarball. As a bonus, `hatch shell` drops us into a virtual environment with the package locally installed, in an editable fashion: we can just edit the files and can immediately test the code without having to reinstall the wheel somewhere else. So, we get the benefits of testing exactly the wheel that is shipped to the users without having to have a separate virtual environment to install the wheel in, without having to manually manage that.

The wheel that gets built is minimal. Here's the corresponding tarball's contents (output of `tar tf`):

```
model_signing-0.0.1a0/model_signing/__init__.py
model_signing-0.0.1a0/model_signing/hashing/__init__.py
model_signing-0.0.1a0/model_signing/hashing/file.py
model_signing-0.0.1a0/model_signing/hashing/hashing.py
model_signing-0.0.1a0/model_signing/hashing/memory.py
model_signing-0.0.1a0/model_signing/manifest/__init__.py
model_signing-0.0.1a0/model_signing/manifest/manifest.py
model_signing-0.0.1a0/model_signing/serialization/__init__.py
model_signing-0.0.1a0/model_signing/serialization/serialization.py
model_signing-0.0.1a0/model_signing/serialization/serialize_by_file.py
model_signing-0.0.1a0/model_signing/serialization/serialize_by_file_shard.py
model_signing-0.0.1a0/model_signing/signature/__init__.py
model_signing-0.0.1a0/model_signing/signature/encoding.py
model_signing-0.0.1a0/model_signing/signature/fake.py
model_signing-0.0.1a0/model_signing/signature/key.py
model_signing-0.0.1a0/model_signing/signature/pki.py
model_signing-0.0.1a0/model_signing/signature/signing.py
model_signing-0.0.1a0/model_signing/signature/sigstore.py
model_signing-0.0.1a0/model_signing/signature/verifying.py
model_signing-0.0.1a0/model_signing/signing/__init__.py
model_signing-0.0.1a0/model_signing/signing/as_bytes.py
model_signing-0.0.1a0/model_signing/signing/empty_signing.py
model_signing-0.0.1a0/model_signing/signing/in_toto.py
model_signing-0.0.1a0/model_signing/signing/signing.py
model_signing-0.0.1a0/model_signing/signing/sigstore.py
model_signing-0.0.1a0/.gitignore
model_signing-0.0.1a0/LICENSE
model_signing-0.0.1a0/README.model_signing.md
model_signing-0.0.1a0/pyproject.toml
model_signing-0.0.1a0/PKG-INFO
```

---

I added a few more formatting settings and run `hatch fmt` across the entire code, affecting `slsa_for_models` too. Added some excludes and we'll work on minimizing those in the future.

---

Updated `.gitignore` to exclude all the artifacts generated during development. Updated GitHub Actions to run via `hatch`. Removed GHA workflows which became obsolete, as well as the requirements files for `model_signing`, since now dependencies are handled via `hatch`.

---

Since CodeQL was complaining about `signature/sigstore.py` and we already have `singing/sigstore.py` with the implementation of signing with Sigstore, I removed the file in `signature`. Incidentally, we might want to rename `signature/` as `pki/` as it only has the parts needed for signing with own PKI. We would then offer this as a feature of the wheel (`pip install model_signing[pki]`) instead of having it be there by default. But these changes will happen in future PRs.

---

This would require rebase on #279, #240 given that we have moved files around. It closes #192 given that all tests have been migrated. It closes #68 due to the same. Similarly, closes #182. Also, closes #231 since we have fixed `slsa_for_models` code to pass `hatch fmt`

#### Release Note
NONE (will add when ready to publish, also with CHANGELOG.md)

#### Documentation
NONE (will add when ready to publish, there's some TODO now to merge `README.md` files)


mihaimaruseac:
> Can we exclude the `.gitignore`?

Yes, I somehow thought that that is the default, but it's not. Will rebase.

spencerschrock:
> > Can we exclude the `.gitignore`?
> 
> Yes, I somehow thought that that is the default, but it's not. Will amend + force push/rebase.

The answer is probably no:

https://hatch.pypa.io/latest/plugins/builder/sdist/#default-file-selection
> The following files are always included and cannot be excluded:
> 
> /pyproject.toml
> /hatch.toml
> /hatch_build.py
> /.gitignore or /.hgignore
> Any defined [readme](https://hatch.pypa.io/latest/config/metadata/#readme) file
> All defined [license-files](https://hatch.pypa.io/latest/config/metadata/#license)

mihaimaruseac:
I was just going to paste that. Yeah, cannot be removed from sdits

src/model_signing/signature/sigstore.py
```diff
@@ -36,19 +36,17 @@
...
-            id_provider: str | None = None
-        ):
+        self, disable_ambient: bool = True, id_provider: str | None = None
+    ):
         token = self.__get_identity_token(disable_ambient, id_provider)
         if not token:
             raise ValueError("No identity token supplied or detected!")
         log.info(
-            f"Signing identity provider: {token.expected_certificate_subject}")
+            f"Signing identity provider: {token.expected_certificate_subject}"
```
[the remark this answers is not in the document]
mihaimaruseac: :tada: good catch

pyproject.toml
```diff
@@ -1,3 +1,60 @@
...
+license = { file = "LICENSE" }
+authors = [
+  { name = "Sigstore Authors", email = "sigstore-dev@googlegroups.com" }
+]
+classifiers = [
+  "License :: OSI Approved :: Apache Software License",
+  "Programming Language :: Python :: 3 :: Only",
+  "Programming Language :: Python :: 3",
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
```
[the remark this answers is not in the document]
mihaimaruseac: I made pytype and linting use 3.11, but for testing we can use 3.9 and 3.10.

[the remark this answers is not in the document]
mihaimaruseac: Turns out we need to be 3.11 minimum. `Self` is introduced in 3.11: https://peps.python.org/pep-0673/

model_signing/signature/sigstore.py
```diff
@@ -1,106 +0,0 @@
...
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-"""This package provides the functionality to sign and verify models
-with sigstore."""
-from typing import Optional
-
-from absl import logging as log
```
spencerschrock, line 18: was deleting this file intentional?

pyproject.toml
```diff
@@ -0,0 +1,106 @@
+[build-system]
+requires = ["hatchling>=1.25"]
+build-backend = "hatchling.build"
+
+[project]
+name = "model_signing"
```
spencerschrock, line 6: the pypi link redirects to "model-signing". Should we make that change?

pyproject.toml
```diff
@@ -0,0 +1,106 @@
...
+  "Topic :: Security :: Cryptography",
+  "Topic :: Scientific/Engineering :: Artificial Intelligence",
+  "Typing :: Typed",
+]
+dependencies = [
+  "cryptography",  # TODO: make pki a feature
+  "in-toto-attestation",
+  "sigstore",
+  "typing_extensions",
+]
```
spencerschrock, line 32: this isn't a match from the old `requirements.in`?

pyproject.toml
```diff
@@ -0,0 +1,106 @@
...
+[tool.ruff]
+line-length = 80
+
+[tool.ruff.format]
+docstring-code-format = true
+line-ending = "lf"
+skip-magic-trailing-comma = true
+
+[tool.ruff.lint]
+select = ["B", "D", "E", "F", "I", "N", "PLC", "PLE", "PT", "SIM", "W"]
```
spencerschrock, line 89: should this be `extend-select`?
https://hatch.pypa.io/1.9/config/static-analysis/#extending-config

Or should we be explicit through `select`?

src/model_signing/signature/pki.py
```diff
@@ -147,28 +158,30 @@ def verify(self, bundle: bundle_pb.Bundle) -> None:
...
         usage = signing_cert_crypto.extensions.get_extension_for_class(
-            x509.KeyUsage)
+            x509.KeyUsage
+        )
         if not usage.value.digital_signature:
             raise VerificationError(
-                ('the certificate is not valid for digital'
-                 ' signature usage'))
+                ("the certificate is not valid for digital" " signature usage")
+            )
```
spencerschrock, line 175: this auto format looks weird now that it fits all on one line

src/model_signing/signature/pki.py
```diff
@@ -147,28 +158,30 @@ def verify(self, bundle: bundle_pb.Bundle) -> None:
...
         ext_usage = signing_cert_crypto.extensions.get_extension_for_class(
-            x509.ExtendedKeyUsage)
+            x509.ExtendedKeyUsage
+        )
         if crypto_oid.ExtendedKeyUsageOID.CODE_SIGNING not in ext_usage.value:
             raise VerificationError(
-                ('the certificate is not valid for code'
-                 ' signing usage'))
+                ("the certificate is not valid for code" " signing usage")
```
spencerschrock, line 181: same

.github/workflows/unit_tests.yml
```diff
@@ -15,40 +29,16 @@ defaults:
...
-    - name: Run unit tests
-      run: |
-        set -euo pipefail
-        .github/workflows/scripts/venv_activate.sh
-        # NOTE: option --full-trace may be useful for troubleshooting.
-        # TODO(#68): Remove the need to create this folder.
-        mkdir testdata
-        pytest -v .
+    - name: Set up Hatch
+      uses: pypa/hatch@a3c83ab3d481fbc2dc91dd0088628817488dd1d5
```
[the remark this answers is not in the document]
mihaimaruseac: Yeah, we have to keep an eye on this and revisit in the future. There's an alternative path to install via `pipx`

model_signing/signature/sigstore.py
mihaimaruseac: Yes. There's `signing/sigstore.py` that signs using the API we're building. This one was flagged when I changed `absl.logging` to just Python `logging` (to minimize dependencies): CodeQL reported that we were logging private details from the sigstore certificate. So, since the functionality was duplicated, I just deleted the file

pyproject.toml
mihaimaruseac: It shouldn't matter, https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#name says that `_`, `-`, and `.` are equivalent

pyproject.toml
mihaimaruseac: It isn't. These are just the dependencies needed to make sure the wheel can be installed and used.

`requirements.in` should be used for development scenarios. We were splitting them into requirements for the CLI POCs and for requirements for testing, but we don't need that

pyproject.toml
di: This does affect how the project name will be displayed at https://pypi.org/project/model-signing/, however.

pyproject.toml
mihaimaruseac: We'll switch to `extend-select` once we minimize the rules we exclude, but right now we have a superset of what's by default, so `select` seems better.

src/model_signing/signature/pki.py
mihaimaruseac: Good catch. I'll fix this

spencerschrock:
also can i selfishly request this patch so the VSCode pytest integration still works with the new layout? As an added bonus it restores the ability to do `pytest .` to run the tests (for anyone with muscle memory).
```diff
diff --git a/pyproject.toml b/pyproject.toml
index 7b02f97..7691c99 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,6 +62,9 @@ installer = "pip"
 parallel = true
 randomize = true
 
+[tool.pytest.ini_options]
+pythonpath = "src"
+
 [tool.ruff]
 line-length = 80
 target-version = "py311"
 ```

pyproject.toml
mihaimaruseac: Oh, then I'll change to the dash. Thank you!

mihaimaruseac:
(wrong click on the close button, sorry)

spencerschrock (approved):
Similarly this should be able to close #231 ?

mihaimaruseac:
Ohh, good catch. Yes!

PR #287, opened 2024-08-15
Configure coverage reporting

mihaimaruseac:
#### Summary
By default, `hatch test -c` coverage report lists only counts of lines and lines missing and percentages but there is no way to see which are the ones that are missing. We don't have an option to generate an html report at the moment (https://github.com/pypa/hatch/pull/1477). Added some options to display missing lines.

Also, by default, all files are included in the report, including tests (covering the test-only code). I removed the tests, but if we decide we should add them that's easy to do.

More importantly, the report lists files that are 100% covered (not useful in CI) and empty files (not useful at all). So, I removed those from the output.

There is another bigger issue that only files that are imported by a test get reported, so if we have code that is not tested at all it will not show up here. We already have such code in `signature/` and `signing/sigstore.py`. Fixing this will be left for later.

Current output is:

```
Name                                   Stmts   Miss  Cover   Missing
--------------------------------------------------------------------
src/model_signing/signing/in_toto.py     168     68    60%   65-78, 181-190, 342-367, 485-512, 660-671, 793-806
--------------------------------------------------------------------
TOTAL                                    743     68    91%
```

Fixing the missing coverage is left for later. We should aim for 95%+ or so coverage, I think.

We should probably make it so that GitHub reports this table back on PRs, so reviewers can quickly ask for more testing without needing to check the GHA report. Punted for later, for now I'll just remember to just keep checking.

Note: I reordered the custom envs in `pyproject.toml` to be in alphabetical order.

#### Release Note
NONE
#### Documentation
NONE

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,6 +63,18 @@ randomize = true
 [[tool.hatch.envs.hatch-test.matrix]]
 python = ["3.11", "3.12"]
 
+[tool.hatch.envs.docs]
+description = """Custom environment for pdoc.
+Use `hatch run docs:serve` to view documentation.
+"""
+extra-dependencies = [
+  "pdoc"
+]
+
+[tool.hatch.envs.docs.scripts]
+serve = "pdoc src/model_signing --docformat google --no-show-source"
+build = "serve --output-directory html"
+
 [tool.hatch.envs.type]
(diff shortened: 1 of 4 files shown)


mihaimaruseac:
Closed this as #291 included the change in it too

PR #291, opened 2024-08-16
Add unit tests for signing with sigstore

mihaimaruseac:
#### Summary
We need to do quite a lot of mocking around Sigstore, but we are able to test all logic in our library. What is left to do for testing is e2e integration tests (https://github.com/sigstore/model-transparency/issues/5) and testing with signing on one OS and verifying on another (https://github.com/sigstore/model-transparency/issues/25). Both of these are integration style tests and we will only be able to run them in GHA. I'll send a PR for those soon.

While testing, I discovered some minor bugs with error reporting and one moderate bug. Fixed in this PR.

We now have achieved 100% test coverage! 🎉

```
Name    Stmts   Miss  Cover   Missing
-------------------------------------
TOTAL     835      0   100%
```

Well, almost. There are 2 files that are not imported by tests at all, so they don't get included in the report:

```
src/model_signing/signature/fake.py
src/model_signing/signature/pki.py
```

This depends on https://github.com/sigstore/model-transparency/pull/287 which configures the coverage reporting.

#### Release Note
NONE
#### Documentation
NONE

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,6 +63,18 @@ randomize = true
 [[tool.hatch.envs.hatch-test.matrix]]
 python = ["3.11", "3.12"]
 
+[tool.hatch.envs.docs]
+description = """Custom environment for pdoc.
+Use `hatch run docs:serve` to view documentation.
+"""
+extra-dependencies = [
+  "pdoc"
+]
+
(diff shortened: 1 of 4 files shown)


haydentherapper (approved):
Nice!

PR #298, opened 2024-08-21
 enable pyupgrade (UP) linter

spencerschrock:
#### Summary

The main rule I'm interested in is [UP035](https://docs.astral.sh/ruff/rules/deprecated-import/), which helps prevent using
deprecated imports. Specifically some of the `typing` imports which now
live in `collections.abc`.

I disabled two rules ([UP012](https://docs.astral.sh/ruff/rules/unnecessary-encode-utf8/), [UP015](https://docs.astral.sh/ruff/rules/redundant-open-modes/)) which consider some optional args as unnecessary,
as there were many occurrences and it seemed opinionated. **But happy to re-add if desired**.

#### Release Note

NONE

#### Documentation

NONE

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,9 +112,13 @@ line-ending = "lf"
 skip-magic-trailing-comma = true
 
 [tool.ruff.lint]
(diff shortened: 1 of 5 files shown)


mihaimaruseac (approved):
Thank you

PR #300, opened 2024-08-27
Add support for Python3.10

mihaimaruseac:
#### Summary
Suport Python 3.10 so that we can use the library in Google Colab without having to reinstall.

We only need 2 changes:
- `Self` needs to be imported from `typing_extensions`
- `hashlib.file_digest` needs to be implemented manually. I backported the implementation that Python uses, adapted to our scenario.

#### Release Note
NONE
#### Documentation
NONE

### DIFF
--- a/src/model_signing/hashing/file.py
+++ b/src/model_signing/hashing/file.py
@@ -48,6 +48,7 @@ Example usage for `OpenedFileHasher`:
 
 import hashlib
 import pathlib
+import sys
 from typing import BinaryIO
 
 from typing_extensions import override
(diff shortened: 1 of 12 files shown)


src/model_signing/hashing/file.py
```diff
@@ -192,11 +193,24 @@ def digest_name(self) -> str:
...
-        # pytype: enable=wrong-arg-types
-        return hashing.Digest(self.digest_name, digest.digest())
+        if sys.version_info >= (3, 11):
+            # https://github.com/python/typeshed/issues/2166
+            # pytype: disable=wrong-arg-types
+            digest = hashlib.file_digest(self._fd, self._algorithm)
+            # pytype: enable=wrong-arg-types
+            return hashing.Digest(self.digest_name, digest.digest())
+        else:
+            # Polyfill for Python 3.10
```
spencerschrock, line 203: TIL about `polyfill`. Just another term for backport?

Optional: The Go dev in me would say drop the `else` to reduce indentation one block, but dont know if this applies in Python so free to ignore.

spencerschrock (dismissed):

src/model_signing/hashing/file.py
mihaimaruseac: It's from Javascript, it's for anything that covers a gap in some implementation (like a browser missing some feature). In our case it's the same as backport.

Good idea on dropping the `else`. I shortened the link too, so now we don't have a line that is too long.

spencerschrock (approved):

PR #301, opened 2024-08-27
Add trusted publishing workflow

mihaimaruseac:
#### Summary
Preparing to add releases to PyPI via Trusted Publishing, via GHA.

There are more steps we can do during the release, but for now they are marked as TODOs. When we get closer to v1 release we will have all of them, but right now they're blocked on having good documentation/release notes support too. Need to see if the following two sections here are used.

PS: Updated the version to the next, to try another release.

#### Release Note
NONE
#### Documentation
NONE

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,59 @@
+# Copyright 2024 The Sigstore Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Release
+
+on:
+  tags:
+    - v*
+
+permissions: {}
+
+jobs:
+  build_wheel:
+    name: Build the wheel
+    if: github.repository == 'sigstore/model-transparency'  # Don't do this in forks
+    runs-on: ubuntu-latest  # Use matrix when we have OS specific pieces
+    steps:
+    - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
+    - name: Set up Hatch
+      uses: pypa/hatch@257e27e51a6a5616ed08a39a408a21c35c9931bc
+    - name: Build artifacts
+      run: hatch build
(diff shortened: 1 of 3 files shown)


haydentherapper (approved):
nice! TODOs SGTM

.github/workflows/release.yml
```diff
@@ -0,0 +1,59 @@
...
+name: Release
+
+on:
+  tags:
+    - v*
+
+permissions: {}
+
+jobs:
+  build_wheel:
```
spencerschrock, line 24: does `build_wheel` need `contents: read` ?

.github/workflows/release.yml
```diff
@@ -0,0 +1,59 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Release
+
+on:
+  tags:
+    - v*
```
spencerschrock, line 19: there's also a `release` trigger which you can further restrict by `published` type if that interest you more than `tags`. Which is what the example workflow at the end of https://docs.pypi.org/trusted-publishers/using-a-publisher/ does.

.github/workflows/release.yml
mihaimaruseac: I thought this is the default, we don't need additional permissions to clone the repo.

.github/workflows/release.yml
mihaimaruseac: Oh, that's a good idea. I sent #302 since `tags` should be under `push`.

Now we have a decision to make: do we want to push a tag and automatically create the release, or create the release and then trigger the workflow for the other steps? I have used either in other repos and can't decide right now

mihaimaruseac: I guess, when in doubt we should follow `sigstore-python`: https://github.com/sigstore/sigstore-python/blob/b5b9e3159c7eca5dd153201fad2c060b09b4d924/.github/workflows/release.yml#L3-L6

Will update #302

.github/workflows/release.yml
spencerschrock: I had always assumed differently, but you're right.

.github/workflows/release.yml
spencerschrock: I've used the tag approach in Go just due to the VCS based immutable nature of modules. But following `sigstore-python` SGTM.

Another aspect to consider is access control. You can protect tags with [tag protection rules](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/configuring-tag-protection-rules). 

I'm not sure there is similar protection for releases?

mihaimaruseac: I think access control might require us to switch to tags? Let's test this on the next release (I'll ping you)

PR #306, opened 2024-09-10
Add scripts for benchmarks using the current API

### LINKED ISSUE #206: Benchmarking scripts
We'd like to have benchmarking scripts to improve on the current shell https://github.com/sigstore/model-transparency/blob/main/model_signing/benchmarks/run.sh

We welcome contributions; this is a really important piece to be able to compare various signing options (serialization, hash + params) and be able to cut a v1 release.

@mihaimaruseac has written some python-based script we can work off.

@font FYI

mihaimaruseac:
#### Summary
These are the scripts to generate sample models for benchmarks and to benchmark serializing models from disk to the in-toto statements (or the digest) used just before signing.

I will follow up with another PR where we have some benchmark data, currently running these. I'll also add those numbers to a README which explains how to run these and what experiments were considering.

Wanted to have it here before I move on cleaning the API a little. If you look at the benchmark harness, we have quite a large number of options and the APIs should be very similar. So I'll work on that next, while benchmarks here are running.

We should be able to use this and compare against the numbers in #13 (and also investigate SHA256 vs BLAKE2). 

Fixes #206 in that the benchmarks are public, but we should still get benchmarks data before releasing stable version of library.

#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/benchmarks/generate.py
+++ b/benchmarks/generate.py
@@ -139,12 +139,12 @@ def add_size_arguments(
         parser: The parser to enhance.
         multiple_files: Whether the generator generates multiple files.
     """
-    parser.add_argument("size", help="Size of the model", type=int)
+    parser.add_argument("size", help="size of the model", type=int)
 
     if multiple_files:
         parser.add_argument(
             "-w",
-            help="Optional weights for for model file sizes to generate",
+            help="optional weights for for model file sizes to generate",
             nargs="+",
             type=int,
         )
@@ -161,11 +161,11 @@ def add_count_arguments(
         parser: The parser to enhance.
         with_dirs: Also add argument to generate the directories.
     """
-    parser.add_argument("-n", help="Number of files", type=int, required=True)
+    parser.add_argument("-n", help="number of files", type=int, required=True)
 
     if with_dirs:
         parser.add_argument(
-            "-m", help="Number of directories", type=int, required=True
+            "-m", help="number of directories", type=int, required=True
         )
 
 
@@ -175,26 +175,26 @@ def add_root_argument(parser: argparse.ArgumentParser) -> None:
     Args:
         parser: The parser to enhance.
     """
-    parser.add_argument("--root", help="Model root path", required=True)
+    parser.add_argument("--root", help="model root path", required=True)
 
 
 def build_parser() -> argparse.ArgumentParser:
     """Builds the command line parser for the generator."""
     parser = argparse.ArgumentParser(
-        description="Generate benchmark data for model signing"
+        description="generate benchmark data for model signing"
     )
     parser.set_defaults(func=generate_file)
     subparsers = parser.add_subparsers(title="Model shapes")
 
     parser_file = subparsers.add_parser(
-        "file", help="Generate all data in a single file (default)"
+        "file", help="generate all data in a single file (default)"
     )
     add_root_argument(parser_file)
     add_size_arguments(parser_file, multiple_files=False)
     parser_file.set_defaults(func=generate_file)
 
     parser_dir = subparsers.add_parser(
-        "dir", help="Generate data split into N files in a single directory"
+        "dir", help="generate data split into N files in a single directory"
     )
     add_root_argument(parser_dir)
     add_size_arguments(parser_dir)
@@ -202,7 +202,7 @@ def build_parser() -> argparse.ArgumentParser:
     parser_dir.set_defaults(func=generate_dir)
 
     parser_matrix = subparsers.add_parser(
-        "matrix", help="Generate data split into N files in M directories"
+        "matrix", help="generate data split into N files in M directories"
     )
     add_root_argument(parser_matrix)
     add_size_arguments(parser_matrix)
@@ -211,7 +211,7 @@ def build_parser() -> argparse.ArgumentParser:
 
     parser_nested = subparsers.add_parser(
         "nested",
-        help="Generate data split into N files in a directory nested M levels",
+        help="generate data split into N files in a directory nested M levels",
     )
     add_root_argument(parser_nested)
     add_size_arguments(parser_nested)
--- a/benchmarks/generate.py
+++ b/benchmarks/generate.py
@@ -17,6 +17,7 @@
 import argparse
 import itertools
 import pathlib
+
 import numpy as np
 
 
@@ -155,7 +156,7 @@ def add_count_arguments(
 ) -> None:
     """Adds the count related arguments to a subparser.
 
-    We have N file. In some cases, we also have M directories.
+    We have N files. In some cases, we also have M directories.
 
     Args:
         parser: The parser to enhance.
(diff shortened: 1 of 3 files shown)


benchmarks/generate.py
```diff
@@ -0,0 +1,228 @@
...
+
+import numpy as np
+
+
+def create_file_of_given_size(path: str, size: int) -> None:
+    """Writes a random file at the given path with given size.
+
+    Args:
+        path: Path to a file to write to. Parents are created if needed.
+        size: Size of data to generate and write to file.
```
jiggoha, line 29: Nit: Size of data in bytes

benchmarks/generate.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+def generate_matrix(args: argparse.Namespace):
+    """Generates a random model as M directories with N files each.
+
+    Args:
+        args: The arguments specifying the request.
+    """
+    sizes = generate_file_sizes(args.size // args.m, args.n, args.w)
+    print(sizes)
```
jiggoha, line 101: Leftover debug?

mihaimaruseac: Leftover debug. Thank you for catching!

benchmarks/generate.py
```diff
@@ -0,0 +1,227 @@
...
+
+    Args:
+        args: The arguments specifying the request.
+    """
+    path = args.root
+    for i in range(args.m):
+        path = f"{path}/d{i}"
+
+    for j, sz in enumerate(generate_file_sizes(args.size, args.n, args.w)):
+        create_file_of_given_size(f"{path}/f{j}", sz)
```
spencerschrock, line 126: this indentation looks wrong, `path` is being set in a loop with no use?

pyproject.toml
```diff
@@ -64,6 +64,22 @@ randomize = true
 [[tool.hatch.envs.hatch-test.matrix]]
 python = ["3.10", "3.11", "3.12"]
 
+[tool.hatch.envs.bench]
+description = """Custom environment for running benchmarks.
+Use `hatch run +py=3... bench:generate ${args}` to generate test models.
+Use `hatch run +py=3... bench:serialize ${args}` to benchmark serialization code.
```
spencerschrock, line 70: out of curiosity whats does the `+py=3...` cover that the python matrix below doesn't?

benchmarks/serialize.py
```diff
@@ -0,0 +1,246 @@
...
+        in_toto_builder(serializer.serialize(args.path))
+
+
+def build_parser() -> argparse.ArgumentParser:
+    """Builds the command line parser for the bechmark runner."""
+    parser = argparse.ArgumentParser(
+        description="Benchmark full serialization of a model for model signing"
+    )
+
+    parser.add_argument("path", help="path to model", type=pathlib.Path)
```
spencerschrock, line 177: your `generate.py` had some args as `required`. It seems `path` may be one we want to make required here as well?

benchmarks/generate.py
mihaimaruseac: The path would be `root/d0/d1/d2/d3/d4/.../d{M}` from the first loop and in the second loop we create files `{path}/f0`, `{path}/f1`, ... `{path}/f{n}`

pyproject.toml
mihaimaruseac: For `hatch run` if you don't select from the matrix it would run in all combinations from the matrix, one after the other. The `+py=3.11`, etc. options allow selecting the python version, so you'd run only once.

benchmarks/generate.py
spencerschrock: Ah, I missed these were nested directories and this was concatenation.

benchmarks/serialize.py
```diff
@@ -0,0 +1,246 @@
...
+        merge_hasher_factory = get_hash_engine_factory(args.merge_hasher)
+        if args.use_shards:
+            serializer = serialize_by_file_shard.DigestSerializer(
+                hasher,
+                merge_hasher_factory(),  # pytype: disable=not-instantiable
+                max_workers=args.max_workers,
+            )
+        else:
+            # This gets complicated because the API here is not matching the
+            # rest. We should fix this.
```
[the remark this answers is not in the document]
mihaimaruseac: This should be step 4 of #172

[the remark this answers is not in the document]
mihaimaruseac: Oh, sorry, I thought this was about the full high-level API. Yes, we probably create an issue for the cleanup on this serializer

benchmarks/serialize.py
```diff
@@ -0,0 +1,246 @@
...
+        if args.use_shards:
+            serializer_factory = serialize_by_file_shard.ManifestSerializer
+        else:
+            serializer_factory = serialize_by_file.ManifestSerializer
+
+        serializer = serializer_factory(hasher, max_workers=args.max_workers)
+
+    # 3. Signing layer
+    if args.skip_manifest:
+        in_toto_builder = id
```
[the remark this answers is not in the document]
mihaimaruseac: Nice catch. I wanted `lambda x: x` which is `id` in Haskell. I'll change

benchmarks/serialize.py
mihaimaruseac: It is required because it does not start with `-`/`--` (it's not an option, it's a positional argument)

benchmarks/serialize.py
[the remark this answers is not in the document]
spencerschrock: I thought "we should fix this" was about the ergonomics of the API not matching the rest of the API.

If "this" relates to the "complicated" code below, I don't think we need the issue.

benchmarks/serialize.py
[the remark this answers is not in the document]
spencerschrock: they're almost identical, I believe the comment was mainly from me having to pause for a second to figure out the scoping, especially since `in_toto_builder` is a function.

benchmarks/serialize.py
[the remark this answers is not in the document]
mihaimaruseac: It's https://github.com/sigstore/model-transparency/blob/fcf9f67af7e52d7d37fca5b3bbda5d3156a94224/src/model_signing/serialization/serialize_by_file.py#L332-L361

benchmarks/serialize.py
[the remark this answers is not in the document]
mihaimaruseac: Yeah, that's why I didn't notice it's not doing what it's supposed to. Desired was a function that returns its argument instead got a function that returns "an address" of the argument. Both are fast and the argument is the expensive part, so no way to notice it

benchmarks/serialize.py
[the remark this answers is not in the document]
mihaimaruseac: I'll leave it to work for #172

spencerschrock (approved):

PR #310, opened 2024-10-03
Add Sigstore sign/verify to CLI

font:
#### Summary

- **Update to support Sigstore sign via CLI**
- **Update to support Sigstore verify via CLI**
- **Update docs for Sigstore sign/verify CLI**
- **Update doc to reflect latest CLI changes**

### DIFF
--- a/src/sign.py
+++ b/src/sign.py
@@ -28,6 +28,7 @@ from model_signing.signature import pki
 from model_signing.signature import signing
 from model_signing.signing import in_toto
 from model_signing.signing import in_toto_signature
+from model_signing.signing import sigstore
 
 
 log = logging.getLogger(__name__)
@@ -54,7 +55,7 @@ def _arguments() -> argparse.Namespace:
     method_cmd = parser.add_subparsers(
         required=True,
         dest="method",
-        help="method to sign the model: [pki, private-key, skip]",
+        help="method to sign the model: [pki, private-key, sigstore, skip]",
     )
     # PKI
     pki = method_cmd.add_parser("pki")
@@ -91,6 +92,16 @@ def _arguments() -> argparse.Namespace:
         type=pathlib.Path,
         dest="key_path",
     )
+    # sigstore
+    sigstore = method_cmd.add_parser("sigstore")
+    sigstore.add_argument(
+        "--use_ambient_credentials",
+        help="use ambient credentials (also known as Workload Identity, default is true)",
+        required=False,
+        type=bool,
+        default=True,
+        dest="use_ambient_credentials",
+    )
     # skip
     method_cmd.add_parser("skip")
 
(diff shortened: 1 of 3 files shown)


mihaimaruseac (changes requested):
I like this. Two more things to do:

- can we make sigstore signing method be the default?
- can you fix the failing lint CI?

font:
@mihaimaruseac Thanks for the review!

> * can we make sigstore signing method be the default?

I don't think we can do this as there doesn't appear to be a way to set a default subparser, and if no method is specified on the command line, that parser's Namespace object does not contain the option attributes defined for it. I may have to look more into seeing if there's a way to do this. Let me know if you have any suggestions.

> * can you fix the failing lint CI?

Done

src/sign.py
```diff
@@ -100,17 +112,27 @@ def _arguments() -> argparse.Namespace:
...
-        return pki.PKISigner.from_path(
+        payload_signer = pki.PKISigner.from_path(
             args.key_path, args.signing_cert_path, args.cert_chain_path
         )
+        return in_toto_signature.IntotoSigner(payload_signer)
+    elif args.method == "sigstore":
+        return sigstore.SigstoreDSSESigner(
+            use_ambient_credentials=args.use_ambient_credentials
+        )
     elif args.method == "skip":
```
font, line 129: @mihaimaruseac Do you know the need to support this `skip` signing option? A cursory search did not reveal where it is used/needed.

src/verify.py
```diff
@@ -73,12 +75,52 @@ def _arguments() -> argparse.Namespace:
...
+        return in_toto_signature.IntotoVerifier(verifier)
+    elif args.method == "pki":
+        _check_pki_flags(args)
+        verifier = pki.PKIVerifier.from_paths(args.root_certs)
+        return in_toto_signature.IntotoVerifier(verifier)
+    elif args.method == "sigstore":
+        return sigstore.SigstoreDSSEVerifier(
+            identity=args.identity, oidc_issuer=args.identity_provider
+        )
+    elif args.method == "skip":
```
font, line 114: @mihaimaruseac Same comment here.

font (commented):
@mihaimaruseac This is ready for re-review.

src/sign.py
mihaimaruseac: I think it got added to see just the hashing part, for debugging? I don't think we need to support it for production.

src/verify.py
mihaimaruseac: I think it got added to see just the hashing part, for debugging? I don't think we need to support it for production.

mihaimaruseac:
> I don't think we can do this as there doesn't appear to be a way to set a default subparser, and if no method is specified on the command line, that parser's Namespace object does not contain the option attributes defined for it. I may have to look more into seeing if there's a way to do this. Let me know if you have any suggestions.

We can use `set_defaults`. I did something similar in the benchmarks: https://github.com/sigstore/model-transparency/blob/9e133c4aa83c133195e270426a4d26b002fd0328/benchmarks/generate.py#L186

src/sign.py
font: Ok, we can leave it in for now and remove when we're closer to production.

font:
> We can use `set_defaults`. I did something similar in the benchmarks:
> 
> https://github.com/sigstore/model-transparency/blob/9e133c4aa83c133195e270426a4d26b002fd0328/benchmarks/generate.py#L186

Thanks @mihaimaruseac!

I had looked at the `set_defaults` approach and even tried it, but it seems it's mostly a good way to organize different functionality into its own code paths.

That is, the `generate.py` code runs into the same problem. Namely that when there's options that are part of a positional argument, those options never get added to that positional argument, or parser's, Namespace object. Here's the specific error:

```
→ python generate.py 
Traceback (most recent call last):
  File "model-transparency/benchmarks/generate.py", line 227, in <module>
    args.func(args)
  File "model-transparency/benchmarks/generate.py", line 81, in generate_file
    create_file_of_given_size(args.root, args.size)
                              ^^^^^^^^^
AttributeError: 'Namespace' object has no attribute 'root'
```

So I'm not sure there's a good way to set a default for a required positional argument that also has its own options unless we just start manipulating the Namespace object directly. But that feels a little kludgy. Thoughts?

mihaimaruseac:
Oh, I only tested with `-h` :(

In this case, let's merge this as it is and then move towards merging the sign and verify CLI scripts into a single one, as then we could make the signer be set via a flag, with defaults.

PR #313, opened 2024-10-14
Add support for Python 3.13.

mihaimaruseac:
#### Summary

Starting from now, we will always support the last 4 versions of Python.  We'll try to support the version used in Colab, that's the one that imposes the lower bound on support.

#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/.github/workflows/unit_tests.yml
+++ b/.github/workflows/unit_tests.yml
@@ -35,7 +35,7 @@ jobs:
       fail-fast: false # Don't cancel other jobs if one fails
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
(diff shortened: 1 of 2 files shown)


spencerschrock (approved):
> Starting from now, we will always support the last 4 versions of Python.

Seems like a good thing to put in `README.model_signing.md`?

mihaimaruseac:
> > Starting from now, we will always support the last 4 versions of Python.
> 
> Seems like a good thing to put in `README.model_signing.md`?

Oh, good idea. Will do when we merge the READMEs (at the latest)

spencerschrock:
I also just noticed this PR was #313, so support for Python 3.13 was added in PR 313 😄

PR #319, opened 2024-11-11
Add explicit identity token argument to `SigstoreSigner`.

spencerschrock:
#### Summary
This matches functionality in `sigstore-python`, allowing callers to provide an identity token provided out-of-band.

https://github.com/sigstore/sigstore-python/blob/cac62e8e611d4286a49ecda7f33d356381ab7919/sigstore/_cli.py#L658-L665

Fixes #316 

#### Release Note
Added an optional `identity_token` argument to Sigstore signer classes, allowing users to specify which OIDC token to use.

PR #320, opened 2024-11-12
Add sigstore CLI flags --staging and --identity-token

spencerschrock:
#### Summary

This is a followup of #319 to allow passing an identity token via the CLI. The staging flag was also added to allow easier testing.

#### Release Note

NONE

#### Documentation

NONE

### DIFF
--- a/src/sign.py
+++ b/src/sign.py
@@ -103,6 +103,21 @@ def _arguments() -> argparse.Namespace:
         default=True,
         dest="use_ambient_credentials",
     )
+    sigstore.add_argument(
+        "--staging",
+        help="Use Sigstore's staging instances, instead of the default"
+        + " production instances",
+        action="store_true",
+        dest="sigstore_staging",
+    )
+    sigstore.add_argument(
+        "--identity-token",
+        help="the OIDC identity token to use",
+        required=False,
+        type=str,
+        default="",
+        dest="identity_token",
+    )
     # skip
     method_cmd.add_parser("skip")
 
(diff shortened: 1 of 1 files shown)


src/sign.py
```diff
@@ -103,6 +103,20 @@ def _arguments() -> argparse.Namespace:
...
+        help="Use Sigstore's staging instances, instead of the default production instances",
+        action="store_true",
+        dest="sigstore_staging",
+    )
+    sigstore.add_argument(
+        "--identity-token",
+        help="the OIDC identity token to use",
+        required=False,
+        type=str,
+        default="",
```
spencerschrock, line 117: `sigstore-python` also supports `SIGSTORE_IDENTITY_TOKEN` and `SIGSTORE_STAGING` for these default values, if that's something we want to match.

src/sign.py
```diff
@@ -103,6 +103,21 @@ def _arguments() -> argparse.Namespace:
         default=True,
         dest="use_ambient_credentials",
     )
+    sigstore.add_argument(
+        "--staging",
+        help="Use Sigstore's staging instances, instead of the default"
+        + " production instances",
```
mihaimaruseac, line 109: Let's use implicit string concatenation instead of `+`. See https://github.com/sigstore/model-transparency/blob/a0fa17360c9d768ea532404eaaf6ef1fb54627dc/src/model_signing/serialization/serialize_by_file.py#L60-L63 for an example.

src/sign.py
```diff
@@ -103,6 +103,21 @@ def _arguments() -> argparse.Namespace:
...
+        action="store_true",
+        dest="sigstore_staging",
+    )
+    sigstore.add_argument(
+        "--identity-token",
+        help="the OIDC identity token to use",
+        required=False,
+        type=str,
+        default="",
```
mihaimaruseac, line 118: Tiny nit: These are the defaults, we can skip adding them

mihaimaruseac (commented):
I'm thinking to avoid the envvars as much as possible, at least in the library parts. It makes for harder to debug behavior

src/sign.py
spencerschrock: Done. Was copying style from the argument above.

mihaimaruseac (approved):

mihaimaruseac:
Thank you

PR #323, opened 2024-11-18
Add a higher level API.

mihaimaruseac:
#### Summary
Currently, this only supports Sigstore (and only DSSE on verification). I want to push this now so that people can start playing with it while I update the Colab demo and try to make the API more uniform and add support for the missing cases.

With this, the default signing should be as simple as

```python
from model_signing import api as model_signing_api

model_signing_api.sign(model_path, signature_path)
```

And verification would be as simple as:

```py
from model_signing import api as model_signing_api

model_signing_api.verify(
  model_path, signature_path,
  expected_identity, expected_oidc_provider
)
```

Will think on how to simplify this even further..

This is part of #172, but not yet closing it

#### Release Note
NONE
#### Documentation
NONE

### DIFF
--- a/src/model_signing/api.py
+++ b/src/model_signing/api.py
@@ -58,7 +58,7 @@ def hash(model_path: os.PathLike) -> manifest.Manifest:
         model_path: the path to the model to hash.
 
     Returns:
-        A manifest that represents hashes for the model.
+        A manifest of the hashed model.
     """
     return HashingConfig().hash(model_path)
 
@@ -68,7 +68,7 @@ def sign(model_path: os.PathLike, signature_path: os.PathLike):
 
     Args:
         model_path: the path to the model to sign.
-        signature_path: the path where to store the resulting signature.
+        signature_path: the path of the resulting signature.
     """
     SigningConfig().sign(model_path, signature_path)
 
@@ -124,12 +124,12 @@ class HashingConfig:
 
         The default hashing configuration uses SHA256 to compute the digest of
         every file in the model. The resulting manifest is a listing of files
-        paired with their hashes. By default, no file is ignored and symbolic
-        links are replaced by their contents.
+        paired with their hashes. By default, no file is ignored and any
+        symbolic link in the model directory results in an error.
         """
         self._ignored_paths = frozenset()
         self._serializer = serialize_by_file.ManifestSerializer(
-            self._build_file_hasher_factory(), allow_symlinks=True
+            self._build_file_hasher_factory(), allow_symlinks=False
         )
 
     def hash(self, model_path: os.PathLike) -> manifest.Manifest:
@@ -425,7 +425,7 @@ class SigningConfig:
 
         Args:
             model_path: the path to the model to sign.
-            signature_path: the path where to store the resulting signature.
+            signature_path: the path of the resulting signature.
         """
         manifest = self._hashing_config.hash(model_path)
         payload = self._payload_generator(manifest)
--- a/src/model_signing/api.py
+++ b/src/model_signing/api.py
@@ -1,0 +1,582 @@
+# Copyright 2024 The Sigstore Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Public, high-level API for the model_signing library.
+
+Users should use this API to sign models and verify the model integrity instead
+of reaching out to the internals of the library. We guarantee backwards
+compatibility only for the API defined in this file.
+"""
+
+from collections.abc import Callable, Iterable
+import os
+import pathlib
+import sys
+from typing import Literal, cast
+
+from model_signing.hashing import file
+from model_signing.hashing import hashing
+from model_signing.hashing import memory
+from model_signing.manifest import manifest
+from model_signing.serialization import serialize_by_file
+from model_signing.serialization import serialize_by_file_shard
+from model_signing.signing import in_toto
+from model_signing.signing import signing
+from model_signing.signing import sigstore
+
+
+if sys.version_info >= (3, 11):
+    from typing import Self
+else:
+    from typing_extensions import Self
+
+
+def hash(model_path: os.PathLike) -> manifest.Manifest:
+    """Hashes a model using the default configuration.
+
+    We use a separate method and configuration for hashing as it needs to be
+    common between signing and signature verification. Having thise separate
+    also helps with performance testing, as hashing is expected to take the
+    largest amount of time (proportional to model size).
+
+    Since we need to be flexible on the serialization format, this returns a
+    manifest, instead of just a single digest. The type of returned manifest
+    depends on the configuration.
+
+    Args:
+        model_path: the path to the model to hash.
+
+    Returns:
+        A manifest that represents hashes for the model.
+    """
+    return HashingConfig().hash(model_path)
+
+
+def sign(model_path: os.PathLike, signature_path: os.PathLike):
+    """Signs a model using the default configuration.
+
+    Args:
+        model_path: the path to the model to sign.
+        signature_path: the path where to store the resulting signature.
+    """
+    SigningConfig().sign(model_path, signature_path)
+
+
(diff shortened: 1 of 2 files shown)


src/model_signing/api.py
```diff
@@ -0,0 +1,582 @@
...
+            max_workers=max_workers,
+            allow_symlinks=allow_symlinks,
+        )
+        return self
+
+    def set_serialize_by_file_to_digest(
+        self,
+        *,
+        hashing_algorithm: Literal["sha256", "blake2"] = "sha256",
+        merge_algorithm: Literal["sha256", "blake2"] = "sha256",
```
spencerschrock, line 260: would we expect these to differ?

src/model_signing/api.py
```diff
@@ -0,0 +1,582 @@
...
+        return factory
+
+    def set_serialize_by_file_to_manifest(
+        self,
+        *,
+        hashing_algorithm: Literal["sha256", "blake2"] = "sha256",
+        chunk_size: int = 8192,
+        max_workers: int | None = None,
+        allow_symlinks: bool = False,
+    ) -> Self:
```
spencerschrock, line 228: the default in the `HashingConfig` is to allow symlinks, which conflicts with this. 

I think it would be good to align the defaults one way or another

src/model_signing/api.py
```diff
@@ -0,0 +1,582 @@
...
+        return self
+
+    def set_serialize_by_file_to_digest(
+        self,
+        *,
+        hashing_algorithm: Literal["sha256", "blake2"] = "sha256",
+        merge_algorithm: Literal["sha256", "blake2"] = "sha256",
+        chunk_size: int = 8192,
+        allow_symlinks: bool = False,
+    ) -> Self:
```
spencerschrock, line 263: (technically this doesn't match the hashing constructor, but this is also a different manifest type)

src/model_signing/api.py
```diff
@@ -0,0 +1,582 @@
...
+
+    def set_serialize_by_file_shard_to_manifest(
+        self,
+        *,
+        hashing_algorithm: Literal["sha256", "blake2"] = "sha256",
+        chunk_size: int = 8192,
+        shard_size: int = 1000000,
+        max_workers: int | None = None,
+        allow_symlinks: bool = False,
+    ) -> Self:
```
spencerschrock, line 307: just highlighting in case we change the defaults from other comment

src/model_signing/api.py
```diff
@@ -0,0 +1,582 @@
...
+            max_workers=max_workers,
+            allow_symlinks=allow_symlinks,
+        )
+        return self
+
+    def set_serialize_by_file_shard_to_digest(
+        self,
+        *,
+        hashing_algorithm: Literal["sha256", "blake2"] = "sha256",
+        merge_algorithm: Literal["sha256", "blake2"] = "sha256",
```
spencerschrock, line 343: same comment about if these would ever differ

src/model_signing/api.py
```diff
@@ -0,0 +1,582 @@
...
+        chunk_size: int = 8192,
+        shard_size: int = 1000000,
+        max_workers: int | None = None,
+        allow_symlinks: bool = False,
+    ) -> Self:
```
spencerschrock, line 348: same potential discussion

src/model_signing/api.py
```diff
@@ -0,0 +1,582 @@
...
+        identity_token: str | None = None,
+    ) -> Self:
+        """Configures the signing to be performed with Sigstore.
+
+        Only one signer can be configured. Currently, we only support Sigstore
+        in the API, but the CLI supports signing with PKI, BYOK and no signing.
+        We will merge the configurations in a subsequent change.
+
+        Args:
+            sign_dsse: Sign a DSSE statement (if True) or a binary blob.
```
spencerschrock, line 482: do we expect these to be the only two options for sigstore?

src/model_signing/api.py
```diff
@@ -0,0 +1,582 @@
...
+            use_ambient_credentials=False, use_staging=False
+        )
+
+    def sign(self, model_path: os.PathLike, signature_path: os.PathLike):
+        """Signs a model using the current configuration.
+
+        Args:
+            model_path: the path to the model to sign.
+            signature_path: the path where to store the resulting signature.
+        """
```
spencerschrock, line 429: same comment about extra word

spencerschrock (commented):
> Will think on how to simplify this even further.

I think the main thing that stands out to me now is that Signing/Verifying is a high level wrapper that tends to be simple, but the devil is in the _Hashing_ details. The `HashingConfig` seems like a copy of the `serialization` API. 

I assume the aim is to be explicit about what is backwards compatible by semver? Either way that seems like the rough point to me currently.

src/model_signing/api.py
mihaimaruseac: Nice catch. I changed it so even in default configuration symlinks are errors.

src/model_signing/api.py
mihaimaruseac: Ideally, no, but I wanted to make it explicit that there are 2 hashing operations involved.

src/model_signing/api.py
```diff
@@ -0,0 +1,582 @@
...
+):
+    """Verifies that a model conforms to a signature.
+
+    Currently, this assumes signatures over DSSE, using Sigstore. We will add
+    support for more cases in a future change.
+
+    Args:
+        model_path: the path to the model to verify.
+        signature_path: the path to the signature to check.
+        identity: The expected identity that has signed the model.
```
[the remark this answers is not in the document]
mihaimaruseac: That would be interesting, but I think right now the identity has to be exact in `sigstore-python` at least. If further verification policies get added, we can add support for them too

src/model_signing/api.py
```diff
@@ -0,0 +1,582 @@
...
+    return HashingConfig().hash(model_path)
+
+
+def sign(model_path: os.PathLike, signature_path: os.PathLike):
+    """Signs a model using the default configuration.
+
+    Args:
+        model_path: the path to the model to sign.
+        signature_path: the path where to store the resulting signature.
+    """
```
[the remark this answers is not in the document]
mihaimaruseac: Changed it to "the path of the resulting signature."

src/model_signing/api.py
```diff
@@ -0,0 +1,582 @@
...
+
+    Since we need to be flexible on the serialization format, this returns a
+    manifest, instead of just a single digest. The type of returned manifest
+    depends on the configuration.
+
+    Args:
+        model_path: the path to the model to hash.
+
+    Returns:
+        A manifest that represents hashes for the model.
```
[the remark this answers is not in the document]
mihaimaruseac: That's better, updated. Thank you!

mihaimaruseac:
> I think the main thing that stands out to me now is that Signing/Verifying is a high level wrapper that tends to be simple, but the devil is in the _Hashing_ details.

That is true. For most usecases, the config should not be used to change anything from the defaults, but we want to allow flexibility.

> The `HashingConfig` seems like a copy of the `serialization` API.

I'm planning to make it even more uniform in the future, having setters for each field and then a verification method just before the first signing event. But for that I'd first need to make the serialization API more uniform.

> I assume the aim is to be explicit about what is backwards compatible by semver? Either way that seems like the rough point to me currently.

Yup. We can then change anything else in the library and as long as these methods have the same interface no user would be broken. In fact, probably we should make everything else private (but need to check with documentation, so let's do this only after we have all documentation done).

src/model_signing/api.py
spencerschrock: This is a repeat of the other symlink comments, phrased differently, and can be resolved.

I had only mentioned the different manifest types because it made the difference in their behavior slightly less contradictory.

src/model_signing/api.py
spencerschrock: Thanks. As long as the defaults are consistent :)

src/model_signing/api.py
```diff
@@ -0,0 +1,582 @@
...
+from model_signing.signing import sigstore
+
+
+if sys.version_info >= (3, 11):
+    from typing import Self
+else:
+    from typing_extensions import Self
+
+
+def hash(model_path: os.PathLike) -> manifest.Manifest:
```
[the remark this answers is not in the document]
mihaimaruseac: Oh, I think `pathlib.Path` constructor supports anything that implements the `os.PathLike` protocol.

spencerschrock (approved):
> I'm planning to make it even more uniform in the future

As we iterate here on the stable API, what's the convention in PyPI? Anything under 1.0 can expect breaking changes?

mihaimaruseac:
> Anything under 1.0 can expect breaking changes?

Yeah, we'll follow https://semver.org/#semantic-versioning-specification-semver

Thank you for the review!

PR #324, opened 2024-11-19
add sigstore signing integration test

spencerschrock:
#### Summary
With the addition of #319, we can test OIDC signing more easily using Sigstore's [extremely dangerous public OIDC beacon](https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon). 

This is the first integration tests and relates to #5. The change is written such that `hatch test` will continue to only run unit tests. Devs and CI jobs that want to run the integration tests need to use the `integration` marker:
```
hatch test -m integration
```

I'm still playing with the test, particularly around handling of expired tokens.

#### Release Note

NONE

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,0 +1,47 @@
+# Copyright 2024 The Sigstore Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Run integration tests
+on:
+  pull_request:
+    branches: [main]
+    types: [opened, synchronize]
+    paths-ignore:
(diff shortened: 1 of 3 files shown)


spencerschrock:
This could also be moved to the higher API level now that #323 is merged

mihaimaruseac:
> This could also be moved to the higher API level now that #323 is merged

+1, let's migrate both this and the benchmarks to the higher level API

spencerschrock:
I'll note there is a lot of variance in runtime depending on the GitHub action scheduling in the conformance repo, I've seen anything from 4s to 3m. So we could have this as a check on every PR (maybe not required?), or set it up on a schedule.

.github/workflows/integration.yml
```diff
@@ -0,0 +1,47 @@
...
+
+jobs:
+  model-signing-integration-test:
+    name: Signing with Python ${{ matrix.python-version }} on ${{ startsWith(matrix.os, 'macos-') && 'macOS' || startsWith(matrix.os, 'windows-') && 'Windows' || 'Linux' }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false # Don't cancel other jobs if one fails
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
+        python-version: ['3.10', '3.11', '3.12', '3.13']
```
spencerschrock, line 39: can make this a smaller matrix if desired, I know we talked about other OS test properties that may need to be manual. (sign on linux, verify on windows)

mihaimaruseac (approved):
Thank you!

PR #326, opened 2024-11-20
use unique hash engine for each file hasher

spencerschrock:
#### Summary

Previously, the factory could cause workers to share a hash engine, leading to hashes that were wrong or duplicated. This ultimately created invalid manifests, affecting signature creation and verification.

Discovered and tested during #324, which I'll keep separate.

#### Release Note

NONE


mihaimaruseac (approved):
Thank you!

PR #336, opened 2025-01-13
add script to run hash algorithm benchmark

spencerschrock:
#### Summary
Builds upon the work in #306 and starts to define individual experiments. This one is aimed specifically at hashing algorithm.

As far as hashing is concerned, bytes are bytes. By generating our own bytes, we avoid I/O associated with reading models from disk. While we could read actual models into memory, recreating the filesystem seems unecessary for this benchmark.

```shell
$ hatch run +py=3.11 bench:hash --repeat 5 --methods sha256 blake2
algorithm: sha256, size: 1.0 KB, best time: 2.569984644651413e-06s
algorithm: blake2, size: 1.0 KB, best time: 3.1301751732826233e-06s
algorithm: sha256, size: 1.0 MB, best time: 0.0007016910240054131s
algorithm: blake2, size: 1.0 MB, best time: 0.001357788685709238s
algorithm: sha256, size: 512.0 MB, best time: 0.41755866911262274s
algorithm: blake2, size: 512.0 MB, best time: 0.7319729649461806s
algorithm: sha256, size: 1.0 GB, best time: 0.8285107491537929s
algorithm: blake2, size: 1.0 GB, best time: 1.466184071265161s
algorithm: sha256, size: 4.0 GB, best time: 3.375688728876412s
algorithm: blake2, size: 4.0 GB, best time: 5.8439111188054085s
algorithm: sha256, size: 16.0 GB, best time: 14.146526537369937s
algorithm: blake2, size: 16.0 GB, best time: 23.482591222040355s
algorithm: sha256, size: 32.0 GB, best time: 21.890016920864582s
algorithm: blake2, size: 32.0 GB, best time: 42.17644724994898s
```

#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/benchmarks/exp_hash.py
+++ b/benchmarks/exp_hash.py
@@ -52,11 +52,7 @@ def build_parser() -> argparse.ArgumentParser:
     )
 
     parser.add_argument(
-        "--data-sizes",
-        help="hash methods to benchmark",
-        nargs="+",
-        type=int,
-        default=[KB, MB, 512 * MB, GB, 4 * GB, 16 * GB, 32 * GB],
+        "--data-sizes", help="hash methods to benchmark", nargs="+", type=int
     )
 
     return parser
@@ -87,11 +83,23 @@ def _generate_data(size: int) -> bytes:
     return np.random.randint(0, 256, size, dtype=np.uint8).tobytes()
 
 
+def _default_sizes() -> list[int]:
+    """Generates sizes following 1, 2, 5 pattern, useful for log scale."""
+    sizes = []
+    for scale in [KB, MB, GB]:
+        for d in [1, 2, 5, 10, 20, 50, 100, 200, 500]:
+            if scale == GB and d > 20:
+                break
+            sizes.append(d * scale)
+    return sizes
+
+
 if __name__ == "__main__":
     np.random.seed(42)
     args = build_parser().parse_args()
-    data = _generate_data(max(args.data_sizes))
-    for size in args.data_sizes:
+    sizes = args.data_sizes or _default_sizes()
+    data = _generate_data(max(sizes))
+    for size in sizes:
         for algorithm in args.methods:
             hasher = _get_hasher(algorithm)
 
--- a/benchmarks/exp_hash.py
+++ b/benchmarks/exp_hash.py
@@ -122,4 +122,4 @@ if __name__ == "__main__":
             # Grab the min time, as suggested by the docs
             # https://docs.python.org/3/library/timeit.html#timeit.Timer.repeat
             measurement = min(times)
-            print(f"{f'{algorithm}/{size}: ':<{padding}}{measurement}")
+            print(f"{f'{algorithm}/{size}: ':<{padding}}{measurement:10.4f}")
--- a/benchmarks/exp_hash.py
+++ b/benchmarks/exp_hash.py
@@ -16,10 +16,18 @@
 """Script for running a benchmark to pick a hashing algorithm."""
 
 import argparse
-import pathlib
 import timeit
+from typing import Final
 
-import serialize
+import numpy as np
+
+from model_signing.hashing import hashing
+from model_signing.hashing import memory
+
+
+KB: Final[int] = 1024
+MB: Final[int] = 1024 * KB
+GB: Final[int] = 1024 * MB
 
 
 def build_parser() -> argparse.ArgumentParser:
@@ -27,7 +35,6 @@ def build_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser(
         description="hash algorithm benchmark data for model signing"
     )
-    parser.add_argument("path", help="path to model", type=pathlib.Path)
 
     parser.add_argument(
         "--repeat",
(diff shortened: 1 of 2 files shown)


benchmarks/exp_hash.py
```diff
@@ -0,0 +1,110 @@
...
+        type=str,
+        default=["sha256", "blake2"],
+    )
+
+    parser.add_argument(
+        "--data-sizes",
+        help="hash methods to benchmark",
+        nargs="+",
+        type=int,
+        default=[KB, MB, 512 * MB, GB, 4 * GB, 16 * GB, 32 * GB],
```
spencerschrock, line 59: These values are arbitrary for now. Happy to change them now/later

mihaimaruseac: [discussion] Let's use 1,2,5,10,20,50,100,200,500,... instead just of powers of 2, to get values in between. This way, an eventual plot would be more easier to understand (an issue in the old plots)


On log scale,distance between 1 and 2 is the same as between 5 and 10 and just slightly lower than between 2 and 5:

```
log10(10) - log10(5) = 0.301
log10(5) - log10(2) = 0.397
log10(2) - log10(1) = 0.301
```

These are the numbers used in [`autorange`](https://docs.python.org/3/library/timeit.html#timeit.Timer.repeat) too.

These are no longer simple powers of 2 so human readable output would change, but for plotting these sizes would be better.

But if we also want human readable sizes / powers of 2 we need different thresholds here. Let me think more about this

benchmarks/exp_hash.py
```diff
@@ -0,0 +1,110 @@
...
+if __name__ == "__main__":
+    np.random.seed(42)
+    args = build_parser().parse_args()
+    data = _generate_data(max(args.data_sizes))
+    for size in args.data_sizes:
+        for algorithm in args.methods:
+            hasher = _get_hasher(algorithm)
+
+            def hash(hasher=hasher, size=size):
+                hasher.update(data[:size])
```
mihaimaruseac, line 99: Should we reinitialize the hasher too under the measured scope? We can make `_get_hasher` return just the constructor and call it here

benchmarks/exp_hash.py
```diff
@@ -0,0 +1,110 @@
...
+def _generate_data(size: int) -> bytes:
+    if size < 0:
+        raise ValueError("Cannot generate negative bytes")
+    return np.random.randint(0, 256, size, dtype=np.uint8).tobytes()
+
+
+if __name__ == "__main__":
+    np.random.seed(42)
+    args = build_parser().parse_args()
+    data = _generate_data(max(args.data_sizes))
```
mihaimaruseac, line 93: This generates the largest amount of data from the start and can lead to memory pressure (or hide the impact of memory pressure at large sizes as now we're seeing it everywhere).

Let's initialize the data insize the size for loop.

benchmarks/exp_hash.py
```diff
@@ -0,0 +1,110 @@
...
+                hasher.update(data[:size])
+                return hasher.compute()
+
+            times = timeit.repeat(lambda: hash(), number=1, repeat=args.repeat)
+
+            # Grab the min time, as suggested by the docs
+            # https://docs.python.org/3/library/timeit.html#timeit.Timer.repeat
+            print(
+                f"algorithm: {algorithm}, "
+                f"size: {_human_size(size)}, "
```
mihaimaruseac, line 108: I'm thinking that if we want to plot this we would not want human readable outputs to distort the reported numbers. Let's add a flag to control output format.

benchmarks/exp_hash.py
```diff
@@ -0,0 +1,110 @@
...
+    data = _generate_data(max(args.data_sizes))
+    for size in args.data_sizes:
+        for algorithm in args.methods:
+            hasher = _get_hasher(algorithm)
+
+            def hash(hasher=hasher, size=size):
+                hasher.update(data[:size])
+                return hasher.compute()
+
+            times = timeit.repeat(lambda: hash(), number=1, repeat=args.repeat)
```
mihaimaruseac, line 102: [discussion] We seem to only have only one statement to run, with no setup. In this case, does it matter if we run `repeat(1, N)` or `timeit(N)`? The last one is simpler but `repeat` gives us statistics (which we then ignore, taking just the minimum right now).

We could look into `autorange` as that runs until execution passes a threshold and then reports both number of iterations and time. So we could output two columns, one for iterations, one for time per iteration. This way, we get closed to [google/benchmark output format](https://github.com/google/benchmark/blob/main/docs/user_guide.md#output-formats). Then we could use [existing tools](https://github.com/google/benchmark/blob/main/docs/tools.md) to statistically compare differences between runs if we do optimizations. Or, we can compare between Python's versions this way.

Depending on what we choose here we change what output we have (so the comment below regarding stats might become irrelevant)

benchmarks/exp_hash.py
```diff
@@ -0,0 +1,110 @@
...
+
+            # Grab the min time, as suggested by the docs
+            # https://docs.python.org/3/library/timeit.html#timeit.Timer.repeat
+            print(
+                f"algorithm: {algorithm}, "
+                f"size: {_human_size(size)}, "
+                f"best time: {min(times)}s"
+            )
```
mihaimaruseac, line 110: [comment only] I was thinking that we would report both min, max, average, stddev and quartiles. It would be more informative and we could see when some measurement is affected by noise.

But then the note you linked to also has valid points. Maybe we could have a flag for full stats vs just the min? But first let's solve the discussions between `timeit`/`autorange`/`repeat`

mihaimaruseac (commented):
Thank you, it looks great. I have a few comments / discussion starters to make this useful both for humans and machines (plotting, comparing between runs)

benchmarks/exp_hash.py
spencerschrock: > We seem to only have only one statement to run, with no setup.
> In this case, does it matter if we run `repeat(1, N)` or `timeit(N)`? The last one is simpler but `repeat` gives us statistics (which we then ignore, taking just the minimum right now).

If we did our setup through the `setup` argument/call, I think it would make a difference in how many times `setup` is called. But I agree as written now it's just a matter of statistics, and I lean towards `repeat` due to the accessibility of `min`. `timeit(N)` gives the total time for N iterations, so we can only get the average. 

> We could look into autorange as that runs until execution passes a threshold and then reports both number of iterations and time

As far as I know, `autorange` isn't configurable, the threshold is always 0.2s. But we could always [re-implement the logic](https://github.com/python/cpython/blob/b5ee0258bf5bb60a5a5a65c64717853e06b64808/Lib/timeit.py#L210-L229).

benchmarks/exp_hash.py
mihaimaruseac: Let's stay with only `min` here. If later we need more info we can add a flag/refactor

benchmarks/exp_hash.py
mihaimaruseac: Let's have a default range for human output (could be the one listed here) and another one for plots. Can make the default be empty list (I think) to signal that the user has not provided an override and then initialize it after we check for the output format flag.

benchmarks/exp_hash.py
mihaimaruseac: For output, let's do `{algoritm}/{size}: {measurement}` with enough spacing for measurement and left aligning it. This would make it more readable as it is one column for scenario and another for the data.

benchmarks/exp_hash.py
spencerschrock: I just included the plot output for now. We can add a `-h` flag later.

`add_argument` will use a default of `None` if not provided, so I just built the logic around that.

benchmarks/exp_hash.py
spencerschrock: since this is in the inner most loop, it's always a new hasher. I dont think we need to reset anything

benchmarks/exp_hash.py
```diff
@@ -0,0 +1,125 @@
...
+            def hash(hasher=hasher, data=data):
+                hasher.update(data)
+                return hasher.compute()
+
+            times = timeit.repeat(lambda: hash(), number=1, repeat=args.repeat)
+
+            # Grab the min time, as suggested by the docs
+            # https://docs.python.org/3/library/timeit.html#timeit.Timer.repeat
+            measurement = min(times)
+            print(f"{f'{algorithm}/{size}: ':<{padding}}{measurement}")
```
mihaimaruseac, line 125: Sorry, I meant to align the measurement, not the size. So we'd have

```
sha256/1KB    0.6 
sha256/2KB   18.7
sha256/5KB   29.8
sha256/10KB 110.0
```

benchmarks/exp_hash.py
mihaimaruseac: But each call of `hash` from `timeit` just hashes data, it doesn't time the time it takes to init the hasher.

benchmarks/exp_hash.py
mihaimaruseac: (or, direct sizes, not human readable ones)

benchmarks/exp_hash.py
spencerschrock: for human output, I think this makes sense. for plotting, something like csv may be better (which would necessitate some bookkeeping before I print)
```csv
size,sha256,blake2
1024,0.01,0.02
2024,0.02,0.04
```

benchmarks/exp_hash.py
spencerschrock:
```
sha256/1024:            0.0000
sha256/2048:            0.0000
sha256/5120:            0.0000
sha256/10240:           0.0000
sha256/20480:           0.0000
sha256/51200:           0.0000
sha256/102400:          0.0001
sha256/204800:          0.0001
sha256/512000:          0.0003
sha256/1048576:         0.0007
sha256/2097152:         0.0013
sha256/5242880:         0.0033
sha256/10485760:        0.0067
sha256/20971520:        0.0133
sha256/52428800:        0.0332
sha256/104857600:       0.0669
```

mihaimaruseac (approved):

PR #337, opened 2025-01-15
Fix potential import confusion (in user tests).

mihaimaruseac:
#### Summary
We have a file called `sigstore.py` which contains `import sigstore` statements, but these imports are for `sigstore-python` library.

However, under some testing configurations, the `import` statement will try to import the same file.

To test this behavior, create a file `test.py` with the following contents:

```python
import test
print("OK")
```

When running `python test.py` you will see the print statement executed twice:

```console
[...]$ python test.py
OK
OK
```

Or, to see the exact same behavior, in an evironment where `sigstore-python` is installed, create a file `sigstore.py` with this content:

```python
import sigstore
print("OK")
```

When running `python sigstore.py` you will see the print statement executed twice:

```console
[...]$ python sigstore.py
OK
OK
```

Renaming the file to anything else will only print the message once.

This behavior can result in broken CI, due to [name shadowing](https://discuss.python.org/t/warning-when-importing-a-local-module-with-the-same-name-as-a-2nd-or-3rd-party-module/27799).

Note that the PR contains a file rename and 4 import changes.

#### Release Note
Users that imported `model_signing.signing.sigstore` directly instead of using the hihg-level API will now need to import `model_signing.signing.sign_sigstore`. See for example the CLI tools

#### Documentation
NONE


src/model_signing/api.py
```diff
@@ -32,8 +32,8 @@
 from model_signing.serialization import serialize_by_file
 from model_signing.serialization import serialize_by_file_shard
 from model_signing.signing import in_toto
+from model_signing.signing import sign_sigstore as sigstore
 from model_signing.signing import signing
-from model_signing.signing import sigstore
```
spencerschrock, line 36: Does the `from model_signing.signing` prefix avoid the confusion? 

To clarify, is this to help library users to avoid the confusion?

spencerschrock (approved):

src/model_signing/api.py
spencerschrock: Ack. The recursive import wasn't clear from the diff or PR description, but I understand now.

PR #342, opened 2025-01-16
change default chunk size to 1 MB

spencerschrock:
#### Summary
change default chunk size to 1 MB

This value was based on the benchmarks in f0a6e96.

#### Release Note
* Fine-tuned a lower level parameter used when hashing files, resulting in a speed improvement. The exact improvement is platform dependent but we observed improvements of 50-80%.

#### Documentation
NONE

### DIFF
--- a/benchmarks/generate.py
+++ b/benchmarks/generate.py
@@ -31,7 +31,7 @@ def create_file_of_given_size(path: str, size: int) -> None:
     """
     file_path = pathlib.Path(path)
     file_path.parent.mkdir(parents=True, exist_ok=True)
-    chunk_size = 8192
+    chunk_size = 1048576
     num_chunks = size // chunk_size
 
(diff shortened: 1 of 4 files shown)


spencerschrock:
I want to test this on a Windows system as well before marking ready.

benchmarks/generate.py
```diff
@@ -31,7 +31,7 @@ def create_file_of_given_size(path: str, size: int) -> None:
     """
     file_path = pathlib.Path(path)
     file_path.parent.mkdir(parents=True, exist_ok=True)
-    chunk_size = 8192
+    chunk_size = 1048576
```
spencerschrock, line 34: This change wasn't benchmarked (but also isn't required to have good performance), but changed for unity.

mihaimaruseac (approved):
Thank you!

PR #349, opened 2025-01-24
bench: return the intoto payload if requested

spencerschrock:
#### Summary
Instead of throwing the payload away, this will allow further benchmarking for things that need the payload, such as signing.

#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/benchmarks/serialize.py
+++ b/benchmarks/serialize.py
@@ -17,6 +17,7 @@
 import argparse
 from collections.abc import Callable
 import pathlib
+from typing import Optional
 
 from model_signing.hashing import file
 from model_signing.hashing import hashing
@@ -103,7 +104,7 @@ def get_file_hasher_factory(
     return _hasher_factory
 
 
-def run(args: argparse.Namespace) -> None:
+def run(args: argparse.Namespace) -> Optional[in_toto.IntotoPayload]:
     """Performs the benchmark.
 
     Args:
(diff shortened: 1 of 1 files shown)


benchmarks/serialize.py
```diff
@@ -146,29 +147,26 @@ def run(args: argparse.Namespace) -> None:
         serializer = serializer_factory(hasher, max_workers=args.max_workers)
 
     # 3. Signing layer
-    if args.skip_manifest:
-        in_toto_builder = id  # Do nothing, just evaluate the argument
+    if args.single_digest:
+        in_toto_builder = in_toto.SingleDigestIntotoPayload
     else:
```
spencerschrock, line 152: the diff looks a little weird, but this is just removing this block and moving the indents over:
```diff
-if args.skip_manifest:
-        in_toto_builder = id  # Do nothing, just evaluate the argument
-else:
```

I'm not sure the call to `id` was needed and it messed with type checking

mihaimaruseac: That was to not transform the manifest into an in-toto statement. I wanted a function like `lambda x: x` and accidentally used `id` (which is that in Haskell but not in Python). In Python it just returns a number for every objects, related to its memory location, etc., a single identifier.

But it looks like `skip_manifest` is no longer used, so this is good.

mihaimaruseac (approved):

PR #356, opened 2025-02-25
add shard benchmark experiment

spencerschrock:
#### Summary
The current default is 1MB, though previous benchmarks used GB in the past, so vary the shard size roughly in that range. Each benchmark is concerned with hash time, as well as manifest time.

In previous experimentation, the shard size does depend a bit on the model being analyzed, I picked two models, but it may be good to check with models that shard differently.

preliminary data says 2GiB is a good shard size
```
python3 benchmarks/exp_shard.py ~/models/falcon-7b
1048576:         9.9655  7787170
2097152:         5.8016  3895373
5242880:         3.3394  1560636
10485760:        2.5500   784869
20971520:        2.1676   393941
52428800:        1.9733   159605
104857600:       1.9009    81596
209715200:       1.8476    42876
524288000:       1.8293    18960
1073741824:      1.8951    11599
2147483648:      1.6447     7595
5368709120:      3.9575     4735
10737418240:     7.2927     4178
```

```
python3 benchmarks/exp_shard.py ~/models/gemma-7b/
1048576:        23.5757 13253524
2097152:        10.8698  6628968
5242880:         6.0967  2653668
10485760:        4.6714  1333666
20971520:        3.9581   668485
52428800:        3.6173   270200
104857600:       3.4757   137385
209715200:       3.3168    70477
524288000:       3.3041    30887
1073741824:      2.9686    16736
2147483648:      2.6510    10440
5368709120:      4.4327     6309
10737418240:     8.5381     5524
```

#### Release Note
NONE

#### Documentation
NONE

PR #357, opened 2025-02-26
change default shard size to 1GB

spencerschrock:
#### Summary
The values was based on the benchmarks in #356.

The exact speed improvement depends on the size of the models being
serialized and of the individual files in the model. Speed improvements
ranged from a 5% improvement to a 87% improvement.

Manifest size is also influenced by shard size (and the model /
individual file size). Increasing shard size from ~1MB to 1GB decreases
manifest size by 3 orders of magnitude (99.9% reduction).

#### Release Note
* Fine-tuned the shard sized used when hashing files, resulting in a modest speed improvement. The exact improvement is dependent on model size but we observed improvements of 5-87%. Manifest size was also reduced by 99.9%.

#### Documentation
NONE

### DIFF
--- a/benchmarks/serialize.py
+++ b/benchmarks/serialize.py
@@ -198,9 +198,9 @@ def build_parser() -> argparse.ArgumentParser:
     )
     param_groups.add_argument(
         "--shard",
-        help="shard size (default: 1073741824)",
+        help="shard size (default: 1000000000)",
         type=int,
-        default=1_073_741_824,
+        default=1_000_000_000,
     )
 
     shard_group = parser.add_argument_group("Serialization modes")
--- a/benchmarks/serialize.py
+++ b/benchmarks/serialize.py
@@ -198,9 +198,9 @@ def build_parser() -> argparse.ArgumentParser:
     )
     param_groups.add_argument(
         "--shard",
-        help="shard size (default: 1000000)",
+        help="shard size (default: 1073741824)",
         type=int,
-        default=1000000,
+        default=1_073_741_824,
     )
 
     shard_group = parser.add_argument_group("Serialization modes")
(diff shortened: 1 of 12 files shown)


benchmarks/serialize.py
```diff
@@ -198,9 +198,9 @@ def build_parser() -> argparse.ArgumentParser:
     )
     param_groups.add_argument(
         "--shard",
-        help="shard size (default: 1000000)",
+        help="shard size (default: 1073741824)",
         type=int,
-        default=1000000,
+        default=1_073_741_824,
```
spencerschrock, line 203: applies everywhere: 
any thoughts on gigabyte vs gibibyte (to use IEC 80000-13 terms)
1_000_000_000 vs 1_073_741_824?

spencerschrock (commented):
Lots of diminishing returns on speed after 50 MiB shard size, but manifest size still decreases until we reach the approximate size of the normal file-based manifest (but at that point having shards at all is pointless).

Given HuggingFace has a default shard size of 5GB (5_000_000_000), I'm assuming something in the 1-2Gi(?)B range makes sense here. Happy to do either

mihaimaruseac (dismissed):

benchmarks/serialize.py
mihaimaruseac: Either works, up to you if you want to change

spencerschrock: no clear winner in benchmarking, leaving as is

spencerschrock: Although looking at the shard definitions/goldens, probably easier to reason about `1_000_000_000`

mihaimaruseac (approved):

PR #369, opened 2025-03-20
address zizmor findings in CI definitions

bobcallaway:

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -21,15 +21,18 @@ on:
     # tags:
     #   - '*'
 
-permissions:
-  contents: read
+permissions: {}
 
 jobs:
   build:
     if: github.repository == 'sigstore/model-transparency'  # Don't do this in forks
     runs-on: ubuntu-latest
+    permissions:
+      contents: write # needed to upload artifact
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
       - name: Set up Hatch
         uses: pypa/hatch@257e27e51a6a5616ed08a39a408a21c35c9931bc
       - name: build docs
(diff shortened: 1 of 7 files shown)


.github/workflows/docs.yml
```diff
@@ -21,15 +21,18 @@ on:
...
-permissions:
-  contents: read
+permissions: {}
 
 jobs:
   build:
     if: github.repository == 'sigstore/model-transparency'  # Don't do this in forks
     runs-on: ubuntu-latest
+    permissions:
+      contents: write # needed to upload artifact
```
spencerschrock, line 31: we don't currently have `contents: write` and the workflow works just fine

https://github.com/sigstore/model-transparency/actions/workflows/docs.yml

.github/workflows/release.yml
```diff
@@ -26,8 +26,12 @@ jobs:
     name: Build the wheel
     if: github.repository == 'sigstore/model-transparency'  # Don't do this in forks
     runs-on: ubuntu-latest  # Use matrix when we have OS specific pieces
+    permissions:
+      contents: write # needed to upload artifact
```
spencerschrock, line 30: same comment

https://github.com/sigstore/model-transparency/actions/runs/12799002622/job/35684169703

.github/workflows/slsa_for_ml.yml
```diff
@@ -59,9 +59,13 @@ jobs:
       hash-ubuntu-latest: ${{ steps.hash.outputs.hash-ubuntu-latest }}
       # hash-macos-latest: ${{ steps.hash.outputs.hash-macos-latest }}
       # hash-windows-latest: ${{ steps.hash.outputs.hash-windows-latest }}
+    permissions:
+      contents: write # needed to upload artifact
```
spencerschrock, line 63: permission not needed

spencerschrock (commented):
Thanks. 

Bumping top-level permissions down and limiting credentials after checkout LGTM.

I don't think the addition of `contents: write` is needed, as our workflows have lacked this permission previously and work as desired.

spencerschrock (approved):
Thanks. I'll need to wait for Mihai to click merge unless you're able to.

PR #373, opened 2025-03-25
cli: Add support for ignoring paths when signing or verifying

stefanberger:
#### Summary

Add support for ignoring paths that should not be signed or be part of the signature verification. This is useful when for example signing a model while keeping the option to *later on* be able to check it into a git repository where files like .gitattributes or a .git directory may be added (see models in huggingface git repos). In this case a verifier needs to be made aware that git-related files are to be ignored and signature verification may then be achieved like this:

python src/verify.py \\
   --model_path /path/to/model/ \\
   --sig_path /path/to/model/model.sig \\
   --ignore_path /path/to/model/.git \\
   --ignore_path /path/to/model/.gitattributes \\
   sigstore \\
   --identity identity \\
   --identity-provider identity_provider

Example: https://huggingface.co/deepseek-ai/DeepSeek-V3-0324/tree/main

### DIFF
--- a/src/model_signing/_cli.py
+++ b/src/model_signing/_cli.py
@@ -14,6 +14,7 @@
 
 """The main entry-point for the model_signing package."""
 
+from collections.abc import Sequence
 import pathlib
 from typing import Optional
 
@@ -57,6 +58,14 @@ _read_signature_option = click.option(
     help="Location of the signature file to verify.",
 )
 
+# Decorator for the commonly used option to ignore certain paths
+_ignore_paths_option = click.option(
+    "--ignore-paths",
+    metavar="IGNORE_PATHS",
+    multiple=True,
+    type=pathlib.Path,
+    help="File paths to ignore when signing or verifying.",
+)
 
 # Decorator for the commonly used option to set the path to the private key
 # (when using non-Sigstore PKI).
@@ -318,6 +327,7 @@ def _verify() -> None:
 @_verify.command(name="sigstore")
 @_model_path_argument
 @_read_signature_option
+@_ignore_paths_option
 @click.option(
     "--identity",
     type=str,
@@ -335,13 +345,15 @@ def _verify() -> None:
 def _verify_sigstore(
     model_path: pathlib.Path,
     signature: pathlib.Path,
+    ignore_paths: Sequence[pathlib.Path],
     identity: str,
     identity_provider: str,
 ) -> None:
     """Verify using Sigstore (DEFAULT verification method).
 
     Verifies the integrity of model at MODEL_PATH, according to signature from
-    SIGNATURE_PATH (given via `--signature` option).
+    SIGNATURE_PATH (given via `--signature` option). Files in IGNORE_PATHS are
+    ignored.
 
     For Sigstore, we also need to provide an expected identity and identity
     provider for the signature. If these don't match what is provided in the
@@ -351,12 +363,15 @@ def _verify_sigstore(
         identity=identity, oidc_issuer=identity_provider
     )
     signature_contents = sigstore.SigstoreSignature.read(signature)
-    _serialize_and_verify(model_path, verifier, signature_contents, signature)
+    _serialize_and_verify(
+        model_path, verifier, list(ignore_paths), signature_contents, signature
+    )
 
 
 @_verify.command(name="key")
 @_model_path_argument
 @_read_signature_option
+@_ignore_paths_option
 @click.option(
     "--public_key",
     type=pathlib.Path,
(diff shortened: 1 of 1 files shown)


mihaimaruseac:
Thank you for opening the PR. This is indeed what we need at this point. However, I was working on a refactoring of the CLI on #375 and the `sign.py`/`verify.py` are now obsolete. Can I ask you to rebase this on the refactored `__main__.py`, please? Apologies for the extra work.

src/model_signing/_cli.py
```diff
@@ -57,6 +57,14 @@
     help="Location of the signature file to verify.",
 )
 
+# Decorator for the commonly used option to ignore certain paths
+_ignore_paths_option = click.option(
+    "--ignore-paths",
+    metavar="IGNORE_PATHS",
+    multiple=True,
+    default=[],
```
mihaimaruseac, line 67: Let's instead declare the type and remove the default (since it's implicit as an empty list, given it's not required). See ~_private_key_option` below.

src/model_signing/_cli.py
```diff
@@ -339,13 +348,15 @@ def _verify() -> None:
 def _verify_sigstore(
     model_path: pathlib.Path,
     signature: pathlib.Path,
+    ignore_paths: tuple[str],
```
mihaimaruseac, line 397: Should be `list[pathlib.Path]` here and below

mihaimaruseac: To be more pedantic, just a `Sequence` would be enough

src/model_signing/_cli.py
```diff
@@ -355,12 +366,15 @@ def _verify_sigstore(
         identity=identity, oidc_issuer=identity_provider
     )
     signature_contents = sigstore.SigstoreSignature.read(signature)
-    _serialize_and_verify(model_path, verifier, signature_contents, signature)
+    _serialize_and_verify(
+        model_path, verifier, list(ignore_paths), signature_contents, signature
```
mihaimaruseac: If you make it a `list`/`Sequence` in the type then `list` would not be needed here.

src/model_signing/_cli.py
```diff
@@ -443,7 +469,7 @@ def hasher_factory(file_path: pathlib.Path) -> file.FileHasher:
             verifier=verifier,
             model_path=model_path,
             serializer=serializer,
-            ignore_paths=[signature_file],
+            ignore_paths=[signature_file] + ignore_paths,
```
mihaimaruseac, line 542: Let's write this as `ignore_paths + [signature_file]` (or extract to a separate list before). It should be slightly faster

```python
>>> import timeit
>>> timeit.repeat("one+many", setup="one=[1]; many=list(range(1000))")
[1.4256296629901044, 1.4093451339867897, 1.427245083003072, 1.357308656006353, 1.3508969740069006]
>>> timeit.repeat("many+one", setup="one=[1]; many=list(range(1000))")
[1.381321033986751, 1.3717010380059946, 1.3515870239934884, 1.3433384770178236, 1.3438171989982948]
```

src/model_signing/_cli.py
```diff
@@ -66,6 +68,15 @@
     help="File paths to ignore when signing or verifying.",
 )
 
+# Decorator for the commonly used option to ignore git related paths
+_ignore_git_paths_option = click.option(
+    "--ignore-git-paths",
+    type=bool,
+    is_flag=True,
+    default=False,
```
mihaimaruseac, line 76: I think this should be default `True` and since it's a flag users can pass `--ignore-git-paths=False` (or we can also make it support `--no-ignore-git-paths`)

src/model_signing/_cli.py
```diff
@@ -88,6 +99,10 @@
 )
 
 
+def _collect_git_related_files(model_path: pathlib.Path) -> list[str]:
+    return glob.glob(join(model_path, ".git*"))
```
mihaimaruseac, line 103: This could be `model_path.glob("**/.git*")` to go into all directories or `model_path.glob(".git*")` to only use top-level

mihaimaruseac (commented):
It looks really great, thank you! I have some suggestions to use the same API everywhere and slightly better typing and performance.

src/model_signing/_cli.py
mihaimaruseac: So this should be the type of a single argument (`pathlib.Path`), even though it gets repeated.

Click takes repeated arguments and places them in a tuple, but in the rest of the code we use them as a list. To be fully correct, we should change to `Sequence[pathlib.Path]` in the function arguments, but that can be a different PR. Right now, there seems to be an implicit conversion from `tuple` to `list` (most likely because the code does not use any API that exists in only one of these types).

src/model_signing/_cli.py
stefanberger: The `type=list[...]` is unfortunately ignored and I always get `<class 'tuple'>`.

src/model_signing/_cli.py
stefanberger: Thanks, took the former.

src/model_signing/_cli.py
stefanberger: Done.

src/model_signing/_cli.py
```diff
@@ -339,13 +349,15 @@ def _verify() -> None:
 def _verify_sigstore(
     model_path: pathlib.Path,
     signature: pathlib.Path,
+    ignore_paths: Sequence[str],
```
mihaimaruseac, line 396: pathlib.Path instead of str?

mihaimaruseac (commented):
One more nit and then it's done.

We'll need to rebase back on main branch since #375 merged

src/model_signing/_cli.py
stefanberger: Fixed.

stefanberger:
> One more nit and then it's done.
> 
> We'll need to rebase back on main branch since #375 merged

Done.

mihaimaruseac (approved):
Thank you! and sorry for the extra work caused by the concurrent change

Issue #331, opened 2024-12-05
Verification of unknown signature configurations


spencerschrock:
**Description**

Under the (sigstore) API introduced in #323, there's `sign` and `verify`. The defaults work well together assuming the entity signing is the same entity verifying, and that the defaults haven't changed between `sign`/`verify`:
```py
model_signing_api.sign(model_path, signature_path)
model_signing_api.verify(model_path, signature_path, identity=expected_identity)
```

However, verification of the underlying signature requires knowledge of the `HashingConfig` used during signing. Again, if everyone uses the default, or the signing/verifying entities are the same that's fine. But defaults can change over time, and one model you download might be signed differently than another.

For the most flexible verification, API users need to inspect the signature before verifying it and:

1. Determine signature type (currently only sigstore DSSE envelope).
2. Determine manifest type, (all 4 combinations of { file, shard} x {manifest, digest of digests} )
	1. If using file shards, confirm shard size
3. Confirm hashing algorithm (sha or blake)

I don't believe any of these have any trust implications, so the default `verify` function could do a lot of this transparently to the user.

mihaimaruseac:
Solved by #424  and the other listed PRs

PR #377, opened 2025-03-27
Remove artifact signing, keep only DDSE/in-toto.

mihaimaruseac:
#### Summary
This is cleanup to reach a single signature format to ease adoption.

Artifact signing requires knowing the hash in advance before verification and this causes API mismatches. CLI only uses DSSE/in-toto, and it doesn't make sense to support it only for API, especially since we did not find any users for this. Plus, computing a single model hash for this signing is much much slower.

This is somewhat related to #331: we'd like to have as little configurations as possible available before we do the 1.0 release so that we would be required to support fewer migration paths in the future.

Since [we plan to incorporate more ML metadata in the signature bundle](https://github.com/cosai-oasis/ws1-supply-chain/issues/4), it does not make sense to support a signature format that is just a single hash with nothing else.

#### Release Note
NONE

#### Documentation
NONE


spencerschrock (approved):
Thanks.

optional: one more leftover reference:

https://github.com/sigstore/model-transparency/blob/26cedd780b8c813ce1f4b2f788e1a083bd962311/tests/signing/sigstore_test.py#L45-L56

mihaimaruseac:
Oh, that makes things even simpler, nice catch!

spencerschrock (approved):

PR #381, opened 2025-03-28
Split the public API into three modules.

mihaimaruseac:
#### Summary

This simplifies the interface slightly as users can import only what they need (e.g., if they only care about hashing in their script they can just import the hashing API).

```python
import model_signing

model_signing.hash.hash(...)
```

This is just the first stage, I actually want the names to be `hashing`, `signing` and `verification`, but cannot do that at the moment, given we already have `hashing` as a directory. I'll send another PR to rename that to `_hashing` when we control more clearly what is the internal and what is the external API and then will clean this up. But first, I need to make sure the public API supports all of our needs for V1.

#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/src/model_signing/__init__.py
+++ b/src/model_signing/__init__.py
@@ -14,4 +14,12 @@
 
 """For the stable high-level API, see model_signing.api."""
 
+from model_signing import hash
+from model_signing import sign
+from model_signing import verify
+
+
 __version__ = "0.2.0"
+
+
+__all__ = ["hash", "sign", "verify"]
--- a/src/model_signing/verify.py
+++ b/src/model_signing/verify.py
@@ -1,0 +1,134 @@
+# Copyright 2024 The Sigstore Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""High level API for the verification interface of model_signing library.
+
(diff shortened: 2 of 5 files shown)


tests/api_test.py
```diff
@@ -86,14 +87,15 @@ class TestSigstoreSigning:
     def test_sign_and_verify(
         self, sigstore_oidc_beacon_token, sample_model_folder, tmp_path
     ):
-        sc = model_signing_api.SigningConfig()
+        sc = sign.SigningConfig()
```
spencerschrock, line 90: Is reducing repetition as much as a thing in Python as it is in Go?

eg. 
`sign.SigningConfig` vs. `sign.Config`
similar with:
`hash.HashingConfig` vs `hash.Config`

src/model_signing/verify.py
```diff
@@ -0,0 +1,134 @@
...
+    ).verify(model_path, signature_path)
+
+
+class VerificationConfig:
+    """Configuration to use when verifying models against signatures.
+
+    The verification configuration is used to decouple between serialization
+    formats and verification types. Having configured the serialization format,
+    this configuration class supports setting up the verification parameters.
+    These should match the signing one.
```
spencerschrock, line 71: another copy/paste doc fix, nice!

src/model_signing/__init__.py
```diff
@@ -14,4 +14,12 @@
 
 """For the stable high-level API, see model_signing.api."""
```
spencerschrock, line 15: stale

src/model_signing/__init__.py
```diff
@@ -14,4 +14,12 @@
...
 
+from model_signing import hash
+from model_signing import sign
+from model_signing import verify
+
+
 __version__ = "0.2.0"
+
+
+__all__ = ["hash", "sign", "verify"]
```
spencerschrock, line 25: note: this affects what is generated for the documentation site.

Is our plan to only offer documentation for the stable API?

mihaimaruseac: Oh, I haven't checked that. But I think only the public part should be documented, I'll mark the rest as private in a future PR.

PR #388, opened 2025-03-31
Make hashing private and with consistent comments.

mihaimaruseac:
#### Summary
This is first part in controlling what the public API is.

We are still making sure the comments are comprehensive to help in IDEs.

I changed the name of the hashers. For the file hasher, since the result is the same as reading the file and then hashing the contents, I changed the name to just use the content hasher's name (so instead of `file-sha256` it would be `sha256`). This resolves some inconsistencies in serialization goldens.

For the sharded hasher, rather than start with `file` the name is now `<algorithm>-sharded <shard_size>`, which is more informative.

I removed the opened file descriptor hash engine for now. We have not found a use for it, we needed a polyfill to handle Python versions below 3.11.

Had to fix a test: the empty files and directories are now ignored so a model that is an empty file should hash exactly the same as a model that is an empty directory.

#### Release Note
NONE

#### Documentation
NONE


src/model_signing/_hashing/file.py
```diff
@@ -64,8 +51,6 @@ class FileHasher(hashing.HashEngine):
     hashing files, instead of any `HashEngine` instance).
     """
 
-    pass
-
```
spencerschrock, line 68: is this just style?

src/model_signing/_hashing/file.py
```diff
@@ -140,92 +132,17 @@ def compute(self) -> hashing.Digest:
     @property
     @override
     def digest_size(self) -> int:
-        """The size, in bytes, of the digests produced by the engine."""
```
spencerschrock, line 143: the docstring gets pulled from the parent class?

spencerschrock (approved):

src/model_signing/_hashing/file.py
mihaimaruseac: They do. It doesn't make sense to override a docstring for an `@override` method, except when we add something extra.

src/model_signing/_hashing/file.py
mihaimaruseac: Yeah, we don't need a `pass` if we already have a docstring.

PR #387, opened 2025-03-31
README: Adjust examples to use -m model_signing now

stefanberger:
Adjust the examples in the README.model_signing.md to use the new -m model_signing module and its renamed options.

Resolves: #386

PR #389, opened 2025-03-31
Don't ignore md files in CI for model signing.

mihaimaruseac:
#### Summary
These CI tests are required, so we need to make sure they get triggered even for documentation only PRs.

See #387 for a case where they are stuck.

#### Release Note
NONE

#### Documentation
NONE


spencerschrock (approved):
> These CI tests are required, so we need to make sure they get triggered even for documentation only PRs.

There are other workarounds we can do later when we're not blocking a PR.

mihaimaruseac:
Agree, this was the cannon approach

PR #390, opened 2025-03-31
ECKeySigner: Filter keys by supported curves and use their parameters

stefanberger:
Per the PublicKeyDetails enum from sigstore only NIST P256/384/521 are supported via the following:

    PKIX_ECDSA_P256_SHA_256 = 5; // See NIST FIPS 186-4
    PKIX_ECDSA_P384_SHA_384 = 12;
    PKIX_ECDSA_P521_SHA_512 = 13;

Therefore, check for supported curves and raise an error if the user wants to use an unsupported curve. Also, use curve-specific parameters, such as the hash and the above-shown identifier when signing.

Link: https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_common.proto

### DIFF
--- a/README.model_signing.md
+++ b/README.model_signing.md
@@ -81,14 +81,14 @@ $ source .venv/bin/activate
 ```bash
 $ MODEL_PATH='/path/to/your/model'
 $ SIG_PATH='./model.sig'
-$ openssl ecparam -name secp256k1 -genkey -noout -out ec-secp256k1-priv-key.pem
-$ openssl ec -in ec-secp256k1-priv-key.pem -pubout > ec-secp256k1-pub-key.pem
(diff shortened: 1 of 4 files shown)


stefanberger:
This still requires some testing, do draft for now...

mihaimaruseac (dismissed):
Nice! Thank you!

stefanberger:
> Nice! Thank you!

Unfortunately tests are failing because the wrong key type was used. Should I adjust the key?

mihaimaruseac:
Yes, please

stefanberger:
I refactored a bit more since pkcs11 support will use common code in new utils.py immediately.

mihaimaruseac (approved):
Nice! Thank you so much!

PR #391, opened 2025-03-31
Simplify manifest interface

mihaimaruseac:
#### Summary

I'm doing a lot of items here, I can split them into multiple PRs if needed, although that would be slightly complex given the number of changes that all need to happen to have CI pass.

First, I'm making the `manifest/manifest.py` file be just `_manifest.py`. Removes an `__init__.py` and one level of import and marks this API as private too. For releases past v1, when we want to offer ability to partially match against a manifest we can either open up this API or just provide helper functions.

Next, I'm removing `DigestManifest`. It does not make sense to maintain the serializers that compute a single hash for the entire model blob. They are slow. For the cases where `DigestManifest` was used to just wrap around a digest, we can just use the `Digest` object directly.

This removal also means removal of the obsolete `DigestSerializer`s. These are not used since we want to record every component of the model with their hash instead of just one hash. This removal opens up other opportunities to remove dead code from the serialization files, but that would come as a separate PR.

Furthermore, we can remove `SingleDigestIntotoPayload`, as it is no longer needed, and was never used for actual signatures. There are further cleanup opportunities in `signing/...`, left for future PRs.

Next, since now all manifests are a collection of items, rather than have two separate classes we switch to having `Manifest` contain everything and then have classes for the keys and values that can be held inside the manifest.

Finally, I'm making sure that all comments in the manifest file are correct, and consistent.

#### Release Note
NONE

#### Documentation
NONE


mihaimaruseac:
Oh, I was wrong, I forgot we are now computing the single digest in parallel too.

But there's still the UX issue that if we use a single digest here we won't be able to verify only a subset of the files of the model.

src/model_signing/signing/in_toto.py
```diff
@@ -317,10 +204,10 @@ def from_manifest(cls, manifest: manifest_module.Manifest) -> Self:
             An instance of `DigestOfDigestsIntotoPayload`.
 
         Raises:
-            TypeError: If the manifest is not `FileLevelManifest`.
+            TypeError: If the manifest is not `Manifest`.
         """
-        if not isinstance(manifest, manifest_module.FileLevelManifest):
-            raise TypeError("Only FileLevelManifest is supported")
+        if not isinstance(manifest, _manifest.Manifest):
+            raise TypeError("Only Manifest is supported")
```
spencerschrock, line 210: are we losing any guarantees here (and below) in terms of distinguishing file-level and shard-level?

tests/serialization/testdata/serialize_by_file_shard/TestManifestSerializer/symlink_model_file
```diff
@@ -1 +0,0 @@
-.:0:22:3aab065c7181a173b5dd9e9d32a9f79923440b413be1e1ffcdba26a7365f719b
```
spencerschrock, line 1: all the other deleted tests are related to digest serializer. This is related to the manifest serializer?

spencerschrock:
> But there's still the UX issue that if we use a single digest here we won't be able to verify only a subset of the files of the model.

or incremental rehashing

(working through reviewing the whole change still, it's an XL)

mihaimaruseac:
I'll try to separate the next ones so that each commit could be reviewed easier. This is the biggest change, I think

src/model_signing/_manifest.py
```diff
@@ -0,0 +1,265 @@
...
+    def from_str(cls, s: str) -> Self:
+        """Builds a manifest key from the string representation.
+
+        It is guaranteed that for any `key` of a derived class `C` and a (valid)
+        string `s`, the following two round-trip properties hold:
+
+        ```
+        str(C.from_str(s)) == s
+        C.from_str(str(key)) == key
+        ```
```
spencerschrock, line 100: do we test this directly? Or just indirectly since we have other tests (e.g. goldens)

src/model_signing/_manifest.py
```diff
@@ -0,0 +1,265 @@
...
+from model_signing._hashing import hashing
+
+
+if sys.version_info >= (3, 11):
+    from typing import Self
+else:
+    from typing_extensions import Self
+
+
+@dataclasses.dataclass(frozen=True, order=True)
```
spencerschrock, line 50: was the addition of `order` caused by anything in particular?

src/model_signing/_manifest.py
```diff
@@ -0,0 +1,265 @@
...
+    """An object that can be a key for an item in a manifest.
+
+    We need to be able to convert the key to string when we serialize the
+    manifest and to rebuild the object from the serialized representation when
+    we deserialize the manifest from the signature.
+    """
+
+    @classmethod
+    @abc.abstractmethod
+    def from_str(cls, s: str) -> Self:
```
spencerschrock, line 91: Just curious, is there anyway to guarantee the subclass also defines a `__str__`?

src/model_signing/serialization/serialize_by_file.py
```diff
@@ -223,110 +222,19 @@ def serialize(
...
 
         Raises:
             ValueError: The model contains a symbolic link, but the serializer
               was not initialized with `allow_symlinks=True`.
         """
         return cast(
-            manifest.FileLevelManifest,
+            _manifest.Manifest,
             super().serialize(model_path, ignore_paths=ignore_paths),
         )
```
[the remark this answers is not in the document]
mihaimaruseac: I'm removing it in the next PR (where we get quite a lot of simplification), I focused this one on the manifest changes and minimal changes outside of it.

src/model_signing/serialization/serialize_by_file_shard.py
```diff
@@ -238,106 +237,19 @@ def serialize(
...
 
         Raises:
             ValueError: The model contains a symbolic link, but the serializer
               was not initialized with `allow_symlinks=True`.
         """
         return cast(
-            manifest.ShardLevelManifest,
+            _manifest.Manifest,
             super().serialize(model_path, ignore_paths=ignore_paths),
         )
```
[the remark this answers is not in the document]
mihaimaruseac: Same

src/model_signing/_manifest.py
mihaimaruseac: I think that's implicitly declared for all classes. If you don't define it, `__str__` calls `__repr__` and if you don't define that then there's an implicit method. We cannot force a class to redefine, the only thing we can do is to test these via goldens/other tests.

spencerschrock (approved):

mihaimaruseac:
I'll merge this and then clean-up the private in public interface to simplify review of what's changed

PR #393, opened 2025-04-01
Simplify and make serialization private

mihaimaruseac:
#### Summary

Make serialization be a private path, realize we already have some dead code (after removal from #391). Then realize we were using a private function across non-test modules, so we move those to the common import (so now `serialize_by_file` and `serialize_by_file_shard` would be equivalent in terms of import size).

Next, see that we only have one serializer for each type, so it does not make sense to use an abstract class. Eliminate abstract class, then rename the serializer classes and the import paths to make import simpler. As a side effect, importing the file hashers becomes a little bit more complex (`file_hashing` instead of just `file`), but serializers are more important, so it makes sense to go this way.

I tried to make each commit be reviewable by itself.

#### Release Note
NONE

#### Documentation
NONE


spencerschrock (approved):
> I tried to make each commit be reviewable by itself.

Appreciated, thanks.

> As a side effect, importing the file hashers becomes a little bit more complex (file_hashing instead of just file), but serializers are more important, so it makes sense to go this way.

Agreed, like the serializer change, but the `file_hashing` change is a little verbose.

Issue #148, opened 2024-03-22
Support cloud filesystems


mihaimaruseac:
We need to support cloud filesystems to sign models stored in cloud buckets without downloading them locally. This is especially useful if we want to sign large models.

laurentsimon:
I'd like if we could have an API that takes as input a file engine interface. But that seems maybe complicated?

mihaimaruseac:
Was thinking to maybe do something similar to [TF's `GFile` API](https://www.tensorflow.org/api_docs/python/tf/io/gfile/GFile) and then our API will just have code like

```py
with gfile.open(filepath) as f:
  do_something_with(f)
```

If the `filepath` is a local file then the local filesystem would be used, otherwise, the prefix URI scheme would redirect to the corresponding cloud implementation.

But we don't want a dependency on TF, so I'm searching to see if there is another library we can use instead

mihaimaruseac:
I think we should make this a low priority, given there doesn't seem to exist an OSS dependency that doesn't bring in TF at this point.

mihaimaruseac:
We can use [`etils[epath]`](https://github.com/google/etils) for this, it implements the API without needing to also depend on TF

goutamadwant:
I am implementing a bounded version of this: accept `etils.epath` cloud URIs at the model hashing boundary while preserving local `pathlib` behavior; traverse remote directories without recursive globs; stream file contents and use seek/range reads for shards; add opt-in GCS/S3 backend extras; and cover hashing, signing/verifying, and CLI behavior with an in-memory remote filesystem. Keys, certificates, trust configuration, and signature bundles remain local in this scope, and I will not claim live-provider coverage that is not exercised.

goutamadwant:
I opened #657 to address this. @mihaimaruseac, could you take a look when you have a chance?

The implementation allows model payloads to use GCS or S3 URIs while keeping keys, certificates, trust configuration, and signature bundles local. Remote payloads are traversed through `etils.epath`, read in bounded chunks, and use seek/range reads for sharded hashing.

Validation covered Python 3.10 through 3.14 with 208 passing tests per version (1,040 executions total). Cloud-specific coverage uses an isolated in-memory `gs://` backend and includes hashing, signing, verification, tamper detection, ignore paths, BLAKE3, sharded reads, and CLI digest behavior.

Live GCS and S3 credentials were not exercised and the PR has this limitation called out.

mihaimaruseac:
[...]

PR #395, opened 2025-04-01
Reduce verbosity in file hashing

mihaimaruseac:
#### Summary
Based on a quick idea, instead of `file_hashing.*` we can do `io.*`. Now, we no longer have duplication (`file_hashing.FileHasher`) and the import is much shorter, we reduce a few of the function calls that are on 3 lines.

Plus, if we later add back hashing of cloud filesystems (#148), the name is better suited.

#### Release Note
NONE
#### Documentation
NONE


spencerschrock (approved):
> Plus, if we later add back hashing of cloud filesystems (#148), the name is better suited.

Agreed, SGTM

PR #397, opened 2025-04-01
Remove one level of class hierarchy.

mihaimaruseac:
#### Summary
Since we only support DSSE payloads, it does not make sense to create an abstract class that then is suclassed by just the DSSE signer/verifier. It made sense when we were testing support for arbitrary binary payloads, but that has a totally different interface, so we removed it.

#### Release Note
NONE
#### Documentation
NONE


src/model_signing/_cli.py
```diff
@@ -236,7 +236,7 @@ def _sign_sigstore(
     Passing the `--use_staging` flag would use that instance instead of the
     production one.
     """
-    signer = sigstore.SigstoreDSSESigner(
+    signer = sigstore.SigstoreSigner(
```
spencerschrock, line 239: shall we de-dupe again?
```python
signer = sigstore.Signer(
```
```python
verifier = sigstore.Verifier(
```

spencerschrock: Or is there a downside for those that will import via
```python
from sigstore import SigstoreSigner
```

mihaimaruseac: Yup, planning to when merging with the key and cert signer.

mihaimaruseac: > Or is there a downside for those that will import via
> 
> ```python
> from sigstore import SigstoreSigner
> ```

It shouldn't, we recommend import module not symbol everywhere

spencerschrock (dismissed):
LGTM, but merge conflict so will likely need to review again

spencerschrock (approved):

PR #399, opened 2025-04-01
Fold payload classes up the class hierarchy

mihaimaruseac:
#### Summary
After #398 we have only one abstract class for the payload but with a subclass which then has another subclass. To simplify everything and prepare for #331, I'm folding everything into one single type now. Every step of the folding is one individual commit.

I'll change the names to remove some repetition once #331 is done.

I removed some type checks which are no longer needed, but not all of them. Fixed some of the comments, but not all of them. Future work is needed, after #331, likely.

#### Release Note
NONE
#### Documentation
NONE


src/model_signing/signing/in_toto.py
```diff
@@ -167,6 +129,7 @@ class DigestsIntotoPayload(IntotoPayload):
     """
 
     predicate_type: Final[str] = "https://model_signing/Digests/v0.1"
```
spencerschrock: are we still happy with this predicate string?

spencerschrock (approved):

src/model_signing/signing/in_toto.py
mihaimaruseac: I'll change it when we get the new format.

PR #400, opened 2025-04-01
Only run SLSA CI for SLSA related changes.

mihaimaruseac:
#### Summary
Given that we are working 99.9999% on model signing, it does not make sense to keep training models in CI and generating SLSA provenances for them. We will revisit this when we restart work on SLSA for ML.

#### Release Note
NONE
#### Documentation
NONE

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -17,9 +17,6 @@ on:
   pull_request:
     branches: [main]
     types: [opened, synchronize]
-    paths-ignore:
-      - '**/*.md'
-      - '*.md'
   workflow_dispatch:
 
 permissions: {}
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -18,6 +18,9 @@ on:
     branches: [main]
     types: [opened, synchronize]
   workflow_dispatch:
+  paths-ignore:
+     - '**/*.md'
+     - '*.md'
 
 permissions: {}
 
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -17,10 +17,10 @@ on:
   pull_request:
     branches: [main]
     types: [opened, synchronize]
+    paths-ignore:
+      - '**/*.md'
+      - '*.md'
   workflow_dispatch:
-  paths-ignore:
-     - '**/*.md'
-     - '*.md'
 
 permissions: {}
 
--- a/.github/workflows/slsa_for_ml.yml
+++ b/.github/workflows/slsa_for_ml.yml
@@ -35,6 +35,7 @@ on:
       - '*.md'
     paths:
       - 'slsa_for_models/**'
+      - '.github/workflows/slsa_for_ml.yml'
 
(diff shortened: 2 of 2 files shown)


spencerschrock:
> Integration tests are fast and can run all the time.

They can take up to ~2.5m FYI, which is still faster than the SLSA ones.

.github/workflows/slsa_for_ml.yml
```diff
@@ -33,6 +33,13 @@ on:
     paths-ignore:
       - '**/*.md'
       - '*.md'
+      # TODO: We only want to run this workflow only for SLSA changes
+      # For now, just disable it from any other location, but in the future
+      # when we restart the work on SLSA we will need to be more accurate here.
+      - 'src/'
+      - 'tests/'
+      - 'docs/'
+      - 'benchmarks/'
```
spencerschrock, line 42: why not use a `paths:` filter?
```yaml
paths-ignore:
  - '**/*.md'
  - '*.md'
paths:
  - slsa_for_models
```

mihaimaruseac: Nice! TIL, I didn't know about it

mihaimaruseac:
> > Integration tests are fast and can run all the time.
> 
> They can take up to ~2.5m FYI, which is still faster than the SLSA ones.

Oh, I only looked at Linux. Reverted that part

spencerschrock:
> Oh, I only looked at Linux. Reverted that part

It also just depends on the beacon repo

.github/workflows/integration.yml
```diff
@@ -17,10 +17,10 @@ on:
   pull_request:
     branches: [main]
     types: [opened, synchronize]
-    paths-ignore:
-      - '**/*.md'
-      - '*.md'
   workflow_dispatch:
+  paths-ignore:
+     - '**/*.md'
+     - '*.md'
```
spencerschrock, line 23: this got restored to the wrong place and the indentation is wrong

.github/workflows/slsa_for_ml.yml
```diff
@@ -33,6 +33,9 @@ on:
     paths-ignore:
       - '**/*.md'
       - '*.md'
+    paths:
+      - 'slsa_for_models/**'
+      - '.github/workflows/slsa_for_ml.yml'
```
spencerschrock, line 38: nice catch on the workflow itself

spencerschrock (approved):
```
FAILED tests/signing/sigstore_test.py::TestSigstoreSigning::test_sign_identity_token_precedence - sigstore.errors.TUFError: Failed to refresh TUF metadata
```

seems like something we should fix in terms of unit tests relying out to network

mihaimaruseac:
> ```
> FAILED tests/signing/sigstore_test.py::TestSigstoreSigning::test_sign_identity_token_precedence - sigstore.errors.TUFError: Failed to refresh TUF metadata
> ```
> 
> seems like something we should fix in terms of unit tests relying out to network

Yeah, probably I missed something during mocking.

PR #404, opened 2025-04-02
codestyle: Reformat files with 'ruff format'

stefanberger:
#### Summary

Reformat 2 files that are currently reformatted every time I run 'ruff format' locally.

### DIFF
--- a/slsa_for_models/pytorch_cifar10.py
+++ b/slsa_for_models/pytorch_cifar10.py
@@ -139,8 +139,7 @@ def train_model(model, loss, optimizer, train):
             running_loss += loss_score.item()
             if i % batch_size == 0:
                 print(
-                    f"[{epoch}, {i:5d}], "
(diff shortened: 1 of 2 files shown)


mihaimaruseac (approved):
Thank you. It's interesting that `hatch fmt` (which runs `ruff` in the end) did not complain about these.

spencerschrock:
> `hatch fmt` (which runs `ruff` in the end) did not complain about these.

the slsa one I can understand because it's not in the package. but the model signing one, maybe a difference in config from the pyproject.toml?

stefanberger:
> > `hatch fmt` (which runs `ruff` in the end) did not complain about these.
> 
> the slsa one I can understand because it's not in the package. but the model signing one, maybe a difference in config from the pyproject.toml?

I run this here over my code and it always reformatted these two files:
```
~/.local/bin/ruff format --config pyproject.toml
```

PR #405, opened 2025-04-02
cli: add --use_staging option to `verify sigstore`

spencerschrock:
#### Summary
The verify operation also needs to use the staging instance if the sign operation was for the staging instance. Otherwise the verification complains:
```
Verification failed with error: failed to build chain: unable to get local issuer certificate
```

#### Release Note

The `verify sigstore` operation can now use Sigstore's staging instance via the `--use_staging` option.

#### Documentation

NONE

### DIFF
--- a/src/model_signing/_cli.py
+++ b/src/model_signing/_cli.py
@@ -97,6 +97,15 @@ _certificate_root_of_trust_option = click.option(
 )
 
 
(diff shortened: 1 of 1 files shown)


mihaimaruseac (approved):
Thank you! Great catch (and for the typo fix too)

PR #407, opened 2025-04-03
Change to new format for the signing payload

mihaimaruseac:
#### Summary
This intoto payload records the model name (better to see `gemma` than `.`), records a root hash (hash over all the hashes, easier to refer to the model by just one single hash), moves the resources to the predicate (to only have one hash in the subject, the root hash), no longer has an empty predicate (since we moved the resources there).

But, most importantly, the predicate also contains enough information to reverse the serialization used to build the model. This is the important step for #331. It doesn't fully solve it yet, that will come once I'm done with the rest of the refactoring (when I'll also remove name duplications, etc.).

```json
{
  "_type": "https://in-toto.io/Statement/v1",
  "subject": [
    {
      "name": "file_model",
      "digest": {
        "sha256": "57d54248751a51781ac2bed647670df0c45d3926e337021585c34a82384b6573"
      }
    }
  ],
  "predicateType": "https://model_signing/signature/v1.0",
  "predicate": {
    "serialization": {
      "method": "shards",
      "hash_type": "sha256-sharded-8",
      "allow_symlinks": true,
      "shard_size": 8.0
    },
    "resources": [
      {
        "algorithm": "sha256-sharded-8",
        "digest": "a37010c994067764d86540bf479d93b4d0c3bb3955de7b61f951caf2fd0301b0",
        "name": ".:0:8"
      },
      {
        "algorithm": "sha256-sharded-8",
        "digest": "bd762002a3528a27fb9a8822f822b949d3c9ab7e860af33039c9aa70ebbbe682",
        "name": ".:8:16"
      },
      {
        "algorithm": "sha256-sharded-8",
        "digest": "a791e1e893ea4260c77475725101fb4cc6ad85f6340f21f10b239184e318cd21",
        "name": ".:16:22"
      }
    ]
  }
}
```

This change also makes the signature payload extensible, so we can add based on https://github.com/cosai-oasis/ws1-supply-chain/issues/4.

This likely can be reviewed commit by commit, although tests won't pass in the middle of the chain.

#### Release Note
Major change in signing payload. All previously generated signatures need to be redone.
#### Documentation
NONE

### DIFF
--- a/src/model_signing/_serialization/file.py
+++ b/src/model_signing/_serialization/file.py
@@ -103,7 +103,9 @@ class Serializer(serialization.Serializer):
             for future in concurrent.futures.as_completed(futures):
                 manifest_items.append(future.result())
 
-        return manifest.Manifest(manifest_items)
+        return manifest.Manifest(
+            model_path.name, manifest_items
+        )
 
     def _compute_hash(
         self, model_path: pathlib.Path, path: pathlib.Path
--- a/src/model_signing/_serialization/file.py
+++ b/src/model_signing/_serialization/file.py
@@ -59,7 +59,7 @@ class Serializer(serialization.Serializer):
         # Precompute some private values only once by using a mock file hasher.
         # None of the arguments used to build the hasher are used.
         hasher = file_hasher_factory(pathlib.Path())
-        self._serialization = manifest._FileSerialization(
+        self._serialization_description = manifest._FileSerialization(
             hasher.digest_name, self._allow_symlinks
         )
 
@@ -111,7 +111,7 @@ class Serializer(serialization.Serializer):
                 manifest_items.append(future.result())
 
         return manifest.Manifest(
-            model_path.name, manifest_items, self._serialization
+            model_path.name, manifest_items, self._serialization_description
         )
 
     def _compute_hash(
--- a/src/model_signing/_serialization/file.py
+++ b/src/model_signing/_serialization/file.py
@@ -56,6 +56,13 @@ class Serializer(serialization.Serializer):
         self._max_workers = max_workers
         self._allow_symlinks = allow_symlinks
 
+        # Precompute some private values only once by using a mock file hasher.
+        # None of the arguments used to build the hasher are used.
+        hasher = file_hasher_factory(pathlib.Path())
+        self._serialization = manifest._FileSerialization(
+            hasher.digest_name, self._allow_symlinks
+        )
+
     @override
     def serialize(
         self,
@@ -104,7 +111,7 @@ class Serializer(serialization.Serializer):
                 manifest_items.append(future.result())
 
         return manifest.Manifest(
-            model_path.name, manifest_items
+            model_path.name, manifest_items, self._serialization
         )
 
     def _compute_hash(
--- a/src/model_signing/_serialization/file_shard.py
+++ b/src/model_signing/_serialization/file_shard.py
@@ -135,7 +135,9 @@ class Serializer(serialization.Serializer):
             for future in concurrent.futures.as_completed(futures):
                 manifest_items.append(future.result())
 
-        return manifest.Manifest(manifest_items)
+        return manifest.Manifest(
+            model_path.name, manifest_items
+        )
 
     def _get_shards(
         self, path: pathlib.Path
--- a/src/model_signing/_serialization/file_shard.py
+++ b/src/model_signing/_serialization/file_shard.py
@@ -87,6 +87,9 @@ class Serializer(serialization.Serializer):
         # None of the arguments used to build the hasher are used.
         hasher = sharded_hasher_factory(pathlib.Path(), 0, 1)
         self._shard_size = hasher.shard_size
+        self._serialization = manifest._ShardSerialization(
+            hasher.digest_name, self._shard_size, self._allow_symlinks
+        )
 
     @override
     def serialize(
@@ -136,7 +139,7 @@ class Serializer(serialization.Serializer):
                 manifest_items.append(future.result())
 
         return manifest.Manifest(
-            model_path.name, manifest_items
+            model_path.name, manifest_items, self._serialization
         )
 
     def _get_shards(
--- a/src/model_signing/_serialization/file_shard.py
+++ b/src/model_signing/_serialization/file_shard.py
@@ -87,7 +87,7 @@ class Serializer(serialization.Serializer):
         # None of the arguments used to build the hasher are used.
         hasher = sharded_hasher_factory(pathlib.Path(), 0, 1)
         self._shard_size = hasher.shard_size
-        self._serialization = manifest._ShardSerialization(
+        self._serialization_description = manifest._ShardSerialization(
             hasher.digest_name, self._shard_size, self._allow_symlinks
         )
 
@@ -139,7 +139,7 @@ class Serializer(serialization.Serializer):
                 manifest_items.append(future.result())
 
         return manifest.Manifest(
-            model_path.name, manifest_items, self._serialization
+            model_path.name, manifest_items, self._serialization_description
         )
 
     def _get_shards(
--- a/src/model_signing/manifest.py
+++ b/src/model_signing/manifest.py
@@ -35,6 +35,7 @@ from collections.abc import Iterable, Iterator
 import dataclasses
 import pathlib
 import sys
+from typing import Any, Final
 
 from typing_extensions import override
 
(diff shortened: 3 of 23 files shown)


src/model_signing/_serialization/file.py
```diff
@@ -56,6 +56,13 @@ def __init__(
         self._max_workers = max_workers
         self._allow_symlinks = allow_symlinks
 
+        # Precompute some private values only once by using a mock file hasher.
+        # None of the arguments used to build the hasher are used.
+        hasher = file_hasher_factory(pathlib.Path())
+        self._serialization = manifest._FileSerialization(
```
spencerschrock, line 62: optional: The base type being SerializationType provided some info, but when I first this line I wasn't sure what self._serialization was for the Serializer class. perhaps `_serialization_desc` or similar?

src/model_signing/_serialization/file_shard.py
```diff
@@ -87,6 +87,9 @@ def __init__(
         # None of the arguments used to build the hasher are used.
         hasher = sharded_hasher_factory(pathlib.Path(), 0, 1)
         self._shard_size = hasher.shard_size
+        self._serialization = manifest._ShardSerialization(
+            hasher.digest_name, self._shard_size, self._allow_symlinks
+        )
```
spencerschrock, line 92: similar optional around serialization desc/type

src/model_signing/manifest.py
```diff
@@ -245,16 +246,156 @@ def key(self) -> ManifestKey:
         return Shard(self._path, self._start, self._end)
 
 
+class SerializationType(metaclass=abc.ABCMeta):
```
spencerschrock, line 249: its subclasses are private, does this need to be?

I see it's passed into the Manifest initializer, but are we expecting others to do so?

src/model_signing/manifest.py
```diff
@@ -245,16 +246,156 @@ def key(self) -> ManifestKey:
...
+        )
+
+
 class Manifest:
     """Generic manifest file to represent a model."""
 
-    def __init__(self, items: Iterable[ManifestItem]):
+    def __init__(
+        self,
+        model: str,
```
spencerschrock, line 386: i know we usually use pathlib.Path or similar for paths, but at first glance i would wonder if this was a path or not. What do you think about `name`, `model_name` or something like that? I know the docstring below explains it, but just what came to mind.

src/model_signing/manifest.py
```diff
@@ -263,3 +404,24 @@ def resource_descriptors(self) -> Iterator[ResourceDescriptor]:
         """Yields each resource from the manifest, one by one."""
         for item, digest in sorted(self._item_to_digest.items()):
             yield ResourceDescriptor(identifier=str(item), digest=digest)
+
+    @property
+    def model_name(self) -> str:
+        """The name of the model when serialized.
+
+        This is the final component of the path and is only informative.
```
spencerschrock, line 412: should the final component of the path be part of the docstring for the initializer arg?

src/model_signing/manifest.py
```diff
@@ -263,3 +404,24 @@ def resource_descriptors(self) -> Iterator[ResourceDescriptor]:
...
+        Changing the name of the model should still result in the same digests
+        after serialization, it must not invalidate signatures. As a result, two
+        manifests with different model names but with the same resource
+        descriptors will compare equal.
+        """
+        return self._name
+
+    @property
+    def serialization(self) -> dict[str, Any]:
+        """The serialization (and arguments) used to build the manifest.
```
spencerschrock, line 422: I have similar thoughts here regarding the `serialization` name not being informative enough of a name. But here it's a more public of a name.

src/model_signing/manifest.py
```diff
@@ -245,16 +246,156 @@ def key(self) -> ManifestKey:
...
+    use when verifying a signature.
+    """
+
+    @property
+    @abc.abstractmethod
+    def serialization_dict(self) -> dict[str, Any]:
+        """The arguments of the serialization method."""
+
+    @classmethod
+    def from_dict(cls, serialization_dict: dict[str, Any]) -> Self:
```
spencerschrock, line 264: thoughts on `serialization_args` and `from_args` or `serialization_parameters` `from_parameters`? `dict` is conveyed already by the typing, and the class docstring mentions the word parameters

src/model_signing/manifest.py
```diff
@@ -245,16 +246,156 @@ def key(self) -> ManifestKey:
...
+        raise ValueError(f"Unknown serialization type {serialization_type}")
+
+    @classmethod
+    @abc.abstractmethod
+    def _from_dict(cls, serialization_dict: dict[str, Any]) -> Self:
+        """Performs the actual build from `from_dict`."""
+
+    @abc.abstractmethod
+    def new_item(self, name: str, digest: hashing.Digest) -> ManifestItem:
+        """Returns a new manifest item with given name and digest."""
```
spencerschrock, line 282: revisit comment: don't understand the relevance of the method to the class

src/model_signing/manifest.py
```diff
@@ -245,16 +246,156 @@ def key(self) -> ManifestKey:
...
+
+
+class _FileSerialization(SerializationType):
+    method: Final[str] = "files"
+
+    def __init__(self, hash_type: str, allow_symlinks: bool = False):
```
spencerschrock, line 288: does it matter that these are positional?

src/model_signing/manifest.py
```diff
@@ -245,16 +246,156 @@ def key(self) -> ManifestKey:
...
+        )
+
+    @override
+    def new_item(self, name: str, digest: hashing.Digest) -> ManifestItem:
+        path = pathlib.PurePosixPath(name)
+        return FileManifestItem(path=path, digest=digest)
+
+
+class _ShardSerialization(SerializationType):
+    method: Final[str] = "shards"
```
spencerschrock, line 325: do we expect any other type of shards where we would want to distinguish this as a file shard?

src/model_signing/signing/signing.py
```diff
@@ -55,98 +56,85 @@
...
-    model is matched with a digest. Because existing tooling only supports
-    established hashing algorithms, we annotate every subject with the actual
-    hash algorithm used to compute the file digest, and use "sha256" as the
-    algorithm name in the digest itself.
-
-    Example (TODO: needs to be updated):
+    """In-toto payload used to represent a model for signing.
+
+    This payload represents all the object (files, shards, etc.) of the model
+    paired with their hases. It can be seen as a serialization of a manifest.
```
spencerschrock, line 63: "paired with their hases"

hashes mispelled

src/model_signing/signing/signing.py
```diff
@@ -55,98 +56,85 @@
...
+    we are limited on what hashes we can use
+    (https://github.com/sigstore/sigstore-python/issues/1018). Each hash follows
+    the format of a ResourceDescriptor: is an object containing a name for the
+    object, the hashing algorithm, and the digest value. These are recorded in
+    the predicate, as part of the `"resources"` list.
+
+    The subject is a name for the model (taken from the model's directory) and a
+    global digest over all the computed digests. This is SHA256 computed over
+    all the digests, in the order they show up in the predicate (we canonicalize
+    this to be in alphabetical order). This digest can be used to refer to the
```
spencerschrock, line 74: note to self: just based on digests? or also the file header like what digestofdigests was?

spencerschrock: and merge hasher is no longer customizable, always sha256?

src/model_signing/signing/signing.py
```diff
@@ -188,19 +199,46 @@ def manifest_from_payload(
...
+            algorithm = resource["algorithm"]
+            digest_value = resource["digest"]
             digest = hashing.Digest(algorithm, bytes.fromhex(digest_value))
-            item = manifest.FileManifestItem(path=path, digest=digest)
-            items.append(item)
+            hasher.update(digest.digest_value)
+            items.append(serialization.new_item(name, digest))
+
+        obtained_digest = hasher.compute().digest_hex
+        if obtained_digest != expected_digest:
```
spencerschrock, line 235: im guessing this is cheap to compute on just digest of digests, but do we need to verify it's consistent?

tests/manifest_test.py
```diff
@@ -21,6 +21,8 @@
 
 
 class TestFileLevelManifest:
+    _manifest_type = manifest._FileSerialization("unused")
```
spencerschrock, line 24: nit: it's not immediately clear what "unused" is

tests/manifest_test.py
```diff
@@ -30,8 +32,12 @@ def test_insert_order_does_not_matter(self):
         digest2 = hashing.Digest("test", b"efgh")
         item2 = manifest.FileManifestItem(path=path2, digest=digest2)
 
-        manifest1 = manifest.Manifest([item1, item2])
-        manifest2 = manifest.Manifest([item2, item1])
+        manifest1 = manifest.Manifest(
+            "test", [item1, item2], self._manifest_type
+        )
+        manifest2 = manifest.Manifest(
+            "test", [item2, item1], self._manifest_type
```
spencerschrock, line 39: nit: similar ambiguity for "test" throughout

tests/signing/testdata/signing/TestSigningPayload/deep_model_folder_shard
```diff
@@ -0,0 +1,42 @@
...
+      }
+    }
+  ],
+  "predicateType": "https://model_signing/signature/v1.0",
+  "predicate": {
+    "serialization": {
+      "method": "shards",
+      "hash_type": "sha256-sharded-1000000000",
+      "allow_symlinks": true,
+      "shard_size": 1000000000.0
```
spencerschrock, line 17: shard size has to be an integer right? is this just a JSON quirk?

tests/signing/testdata/signing/TestSigningPayload/empty_model_file_shard
```diff
@@ -0,0 +1,21 @@
+{
+  "_type": "https://in-toto.io/Statement/v1",
+  "subject": [
+    {
+      "name": "empty_file_model",
+      "digest": {
+        "sha256": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
+      }
```
spencerschrock, line 8: just to confirm this is the empty sha256 right?

tests/signing/testdata/signing/TestSigningPayload/sample_model_file
```diff
@@ -2,17 +2,25 @@
...
+    "serialization": {
+      "method": "files",
+      "hash_type": "sha256",
+      "allow_symlinks": true
+    },
+    "resources": [
+      {
+        "algorithm": "sha256",
+        "digest": "3aab065c7181a173b5dd9e9d32a9f79923440b413be1e1ffcdba26a7365f719b",
+        "name": "."
```
spencerschrock, line 22: this has to do with the fact that it's a single file model right? And it's relative path to the root? hence the `.` name?

spencerschrock (commented):
In general LGTM. I think the intoto form is improved. I know I complained about naming in some places, specifically around "serialization", but in terms of the intoto statement I think "serialization" is descriptive enough in the statement.

the predicate looks good to me, but do we have anyone with "intoto readability" that we want to glance at the new type?

src/model_signing/_serialization/file_shard.py
mihaimaruseac: Changed in the same way

src/model_signing/manifest.py
mihaimaruseac: I still have a TODO item to go over all the files and add/remove `_` to make the APIs properly public private. For now I  know for sure that the subclasses don't need to be exposed, but the rest not sure yet

src/model_signing/manifest.py
mihaimaruseac: Those are better names, thank you!

src/model_signing/manifest.py
mihaimaruseac: Changed it to be more descriptive. Thanks for catching!

src/model_signing/manifest.py
mihaimaruseac: They cannot be confused since they have different types, but we could add an `*` and make them keyword based if really needed.

src/model_signing/manifest.py
mihaimaruseac: If we do, that would be a separate serializer, with separate manifest item type and separate serialization type. But right now we should only have these two types

src/model_signing/manifest.py
mihaimaruseac: Right, `model_name` sounds much better

src/model_signing/manifest.py
mihaimaruseac: Added

src/model_signing/manifest.py
mihaimaruseac: Changed to `serialization_type`

src/model_signing/signing/signing.py
mihaimaruseac: Nice catch, fixed

src/model_signing/signing/signing.py
mihaimaruseac: Just the digests, since at this point the names are already recorded below so we catch models where the name changes.

This has to be a hash that sigstore accepts (https://github.com/sigstore/sigstore-python/issues/1018), so always using sha256 is better than allowing customization.

This is somewhat what we used to have as `DigestOfDigests` payload

src/model_signing/signing/signing.py
mihaimaruseac: We need to, otherwise someone could change the root digest to match the digest from another document that does not have the full manifest and attach fake properties to the model

tests/manifest_test.py
mihaimaruseac: I'll change

tests/manifest_test.py
mihaimaruseac: I'll change

tests/signing/testdata/signing/TestSigningPayload/deep_model_folder_shard
mihaimaruseac: I think so, though when I tested from python interpreter with `json.dumps` it used int. I'll debug when I get to more testing e2e (which your PR helps with)

tests/signing/testdata/signing/TestSigningPayload/empty_model_file_shard
mihaimaruseac: Yes, it has to be

```
[...]$ echo -n "" | sha256sum 
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855  -
```

tests/signing/testdata/signing/TestSigningPayload/sample_model_file
mihaimaruseac: That's true.

src/model_signing/manifest.py
spencerschrock: From a intoto standpoint I'm wondering about this part. Since the method here would need to be unique?
```
  "predicate": {
    "serialization": {
      "method": "shards",
```

mihaimaruseac: Yes, we will guarantee it to be unique

spencerschrock (approved):

src/model_signing/manifest.py
```diff
@@ -257,29 +257,41 @@ class SerializationType(metaclass=abc.ABCMeta):
 
     @property
     @abc.abstractmethod
-    def serialization_dict(self) -> dict[str, Any]:
+    def serialization_parameters(self) -> dict[str, Any]:
         """The arguments of the serialization method."""
 
     @classmethod
-    def from_dict(cls, serialization_dict: dict[str, Any]) -> Self:
+    def from_args(cls, args: dict[str, Any]) -> Self:
```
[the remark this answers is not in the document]
mihaimaruseac: Oh, that's a good suggestion. I'll do that in a future PR when we also change the other types to more abstract ones.

PR #408, opened 2025-04-03
Rename `SigningPayload` and remove static method

mihaimaruseac:
#### Summary
We don't need to call `signing.SingingPayload.from_manifest(manifest)` when `signing.Payload(manifest)` is simpler<.

Only the last commit is new, the rest are #407.

#### Release Note
NONE
#### Documentation
NONE


tests/signing/sigstore_test.py
```diff
@@ -229,9 +224,8 @@ def test_sign_verify_mocked_prod(
         manifest = serializer.serialize(sample_model_folder)
         signature_path = tmp_path / "model.sig"
         self._sign_manifest(
-            manifest,
+            signing.Payload(manifest),
```
spencerschrock, line 227: rename `_sign_manifest` to `_sign_payload` if the first arg we're passing in is a payload?

spencerschrock: or would it be simpler to keep it `_sign_manifest` and still pass in the manifest.

and the first line of `_sign_manifest` could just be `payload = signing.Payload(manifest)`

spencerschrock (commented):
LGTM, only minor things. Approval pending on #407

tests/signing/sigstore_test.py
mihaimaruseac: Changed to the second one

spencerschrock (approved):

PR #409, opened 2025-04-03
Simplify payload to manifest transformation

mihaimaruseac:
#### Summary
Rather than using the `Payload` class to call a static class method, convert that to a free function. We can do that since we only support one single signature format and the entire signature format is localized within the `signing` and `signature` directories (to be merged).

I'm also extracting the predicate type to a global of the module instead of accessing it via the `Payload` class. It's local to the module, so that's ok.

We'll keep the `Payload` class though since otherwise we'd need to use the in-toto statement import in multiple other places, outside of the `signing`/`signature` directories and their tests.

Only the last 2 commits are new here (from "Free function instead of manifest_from_payload"). The rest is #408

#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/src/model_signing/signing/signing.py
+++ b/src/model_signing/signing/signing.py
@@ -39,7 +39,7 @@ https://docs.sigstore.dev/about/bundle/.
 import abc
 import pathlib
 import sys
-from typing import Any, Final
+from typing import Any
 
 from in_toto_attestation.v1 import statement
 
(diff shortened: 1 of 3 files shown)


src/model_signing/signing/signing.py
```diff
@@ -56,6 +54,63 @@
     from typing_extensions import Self
 
 
+# The expected model signature predicate type.
+_PREDICATE_TYPE: str = "https://model_signing/signature/v1.0"
+
+
+def dsse_payload_to_manifest(dsse_payload: dict[str, Any]) -> manifest.Manifest:
```
spencerschrock, line 61: flagging as possible dict -> Mapping cleanup later

spencerschrock (approved):

PR #410, opened 2025-04-03
Display certificates' fingerprints when verifying

stefanberger:
#### Summary

Add --fingerprints option to 'verify certificate' and pass the boolean
through the API to display the fingerprints of the certificate chain.
The output may now look like this for a certificate chain with a root-CA
and a signing-CA (2nd):

sha256 Fingerprint: 92:DD:49:37:A2:F1:2B:...
sha256 Fingerprint: AC:1C:27:DD:13:DF:7A:...
Verification succeeded


src/model_signing/_cli.py
```diff
@@ -458,12 +458,21 @@ def _verify_private_key(
 @_ignore_paths_option
 @_ignore_git_paths_option
 @_certificate_root_of_trust_option
+@click.option(
+    "--fingerprints",
```
mihaimaruseac, line 462: Can we change this, and the variable and keyword argument to be `show_fingerprints` or `log_fingerprints`? Makes it more informative

src/model_signing/_signing/sign_certificate.py
```diff
@@ -106,7 +109,10 @@ def __init__(
               signature and establish chain of trust. By default this is empty,
               in which case we would use the root certificates from the
               operating system, as per `certifi.where()`.
+            fingerprints: Display the fingerprints of certificates
         """
+        self.fingerprints = fingerprints
```
mihaimaruseac, line 114: I think this can be private to the class, so `self._log_fingerprints` or similar?

src/model_signing/_signing/sign_certificate.py
```diff
@@ -116,8 +122,17 @@ def __init__(
 
         self._store = crypto.X509Store()
         for certificate in certificates:
+            if self.fingerprints:
+                self.display_cert_fingerprint(certificate, hashes.SHA256())
             self._store.add_cert(crypto.X509.from_cryptography(certificate))
 
+    def display_cert_fingerprint(self, cert, hash_alg: hashes.Hash) -> None:
+        fp = cert.fingerprint(hash_alg)
+        print(
```
mihaimaruseac, line 131: I'm really debating if we should use print or just logging. This is in the library, and we'd like to test it at some point.

Would it be too much work to switch to logging instead?

src/model_signing/_signing/sign_certificate.py
```diff
@@ -140,26 +157,28 @@ def _verify_certificates(
...
+                self.display_cert_fingerprint(cert, hashes.SHA256())
+            return crypto.X509.from_cryptography(cert)
 
         signing_chain = verification_material.x509_certificate_chain
         signing_certificate = x509.load_der_x509_certificate(
             signing_chain.certificates[0].raw_bytes
         )
 
-        max_signing_time = signing_certificate.not_valid_before
+        max_signing_time = signing_certificate.not_valid_before_utc
```
mihaimaruseac, line 171: Nice, thanks for catching the bug!

mihaimaruseac (commented):
This looks really good, I only have a main comment about the naming of the flag/variable/argument/private member and then a suggestion to use logging instead of printing.

src/model_signing/_cli.py
stefanberger: Sure, I can change it.

src/model_signing/_signing/sign_certificate.py
```diff
@@ -116,8 +126,16 @@ def __init__(
 
         self._store = crypto.X509Store()
         for certificate in certificates:
+            if self._log_fingerprints:
+                self.display_cert_fingerprint(certificate, hashes.SHA256())
             self._store.add_cert(crypto.X509.from_cryptography(certificate))
 
+    def display_cert_fingerprint(self, cert, hash_alg: hashes.Hash) -> None:
```
mihaimaruseac, line 133: This does not use `self`, it can be a global free function (outside of the class).

My bad, I think I was not clear in previous review when I said to make this private, I meant to make it `_log_cert_fingerprint` (with the leading `_` so it should not be used outside of the module -- except in tests)

stefanberger: Fixed.

mihaimaruseac (approved):
Thank you!

PR #416, opened 2025-04-04
Make the API fully configurable

mihaimaruseac:
#### Summary
The API now supports all signing and verification types.
#### Release Note
NONE
#### Documentation
NONE


src/model_signing/hashing.py
```diff
@@ -162,23 +160,23 @@ def _build_sharded_file_hasher_factory(
...
 
-        def factory(
+        def _factory(
             path: pathlib.Path, start: int, end: int
         ) -> io.ShardedFileHasher:
+            hasher = self._build_stream_hasher(hashing_algorithm)
             return io.ShardedFileHasher(
                 path,
-                algorithm,
+                hasher,
```
spencerschrock, line 170: was this another instance of #326

src/model_signing/verifying.py
```diff
@@ -35,31 +37,14 @@
...
-        oidc_issuer: The expected OpenID Connect issuer that provided the
-          certificate used for the signature.
-        use_staging: Use staging configurations, instead of production. This
-          is supposed to be set to True only when testing. Default is False.
+def verify(model_path: os.PathLike, signature_path: os.PathLike):
+    """Verifies that a model matches a signature in the default configuration.
+
+    There is no default configuration with default arguments that can be used,
+    so this just raises an exception to signal that verification configuration
+    needs to be set up.
```
spencerschrock, line 45: why have the function if it will always raise an exception?

src/model_signing/verifying.py
```diff
@@ -102,19 +90,14 @@ def set_hashing_config(self, hashing_config: hashing.Config) -> Self:
...
         return self
 
-    def set_sigstore_dsse_verifier(
-        self,
-        *,
-        identity: str,
-        oidc_issuer: Optional[str] = None,
-        use_staging: bool = False,
+    def use_sigstore_verifier(
+        self, *, identity: str, oidc_issuer: str, use_staging: bool = False
```
spencerschrock, line 94: issuer no longer optional seems intentional?

src/model_signing/hashing.py
mihaimaruseac: Yup!

src/model_signing/verifying.py
mihaimaruseac: It is, it was a mistake to not have it optional at the start.

src/model_signing/verifying.py
mihaimaruseac: It will still raise an exception if you use the default config without setting a verifier. We don't have enough optional arguments here to set a default verifier

src/model_signing/verifying.py
spencerschrock: From a design standpoint? The test in api_test.py worked for example

mihaimaruseac: Yes, but that's strange. All sigstore verification requires it. We can make it optional later, but making it required when it's optional breaks compatibility.

spencerschrock: Seems to imply optional here.
https://github.com/sigstore/sigstore-python/blob/310371e4fae59555db8007f01e5af3984bac7e57/sigstore/verify/policy.py#L433-L443

spencerschrock (approved):

src/model_signing/verifying.py
mihaimaruseac: Ah, I was looking at https://sigstore.github.io/sigstore-python/verify/#verifying-against-a-bundle where it was mandatory. Probably it got made optional more recently.

PR #419, opened 2025-04-04
Simplify ignoring git paths

mihaimaruseac:
#### Summary
First, move the ignoring of paths into the API, so that the CLI code is a little bit simpler.

Then, realize that we already exclude paths if they are relative to a path in the ignore list, so we only need to add `.git` to exclude.

#### Release Note
NONE

#### Documentation
NONE

### DIFF
--- a/src/model_signing/hashing.py
+++ b/src/model_signing/hashing.py
@@ -56,7 +56,7 @@ def hash(model_path: os.PathLike) -> manifest.Manifest:
     depends on the configuration.
 
     Args:
-        model_path: the path to the model to hash.
+        model_path: The path to the model to hash.
 
     Returns:
         A manifest of the hashed model.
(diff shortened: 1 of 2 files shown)


src/model_signing/hashing.py
```diff
@@ -90,12 +90,17 @@ def __init__(self):
...
         """
         self._ignored_paths = frozenset()
+        self._ignore_git_paths = True
         self.use_file_serialization()
 
     def hash(self, model_path: os.PathLike) -> manifest.Manifest:
         """Hashes a model using the current configuration."""
+        ignored_paths = [path for path in self._ignored_paths]
+        if self._ignore_git_paths:
+            ignored_paths.append(".git/")
```
spencerschrock, line 100: > We already check in `should_ignore` from `serialization.py` if a path is
relative to an ignored path, so that should already exclude the git
paths if `.git` is added to the excludes.

Isn't this slightly different though? As written, this ignores the top-level `.git` directory. Previously you ignored every `.git` and `.gitattributes`. I'm not saying that was better, just different

spencerschrock (commented):
one comment, otherwise LGTM

src/model_signing/hashing.py
mihaimaruseac: Yes, but we actually want to only ignore the `.git` directory from the model root. It's very unlikely that there will be another `.git` directory somewhere deep inside the model.

Re `.gitattributes`, I'll add it.

spencerschrock: personally I think it was better just as `.git` and not the other files, I was mainly pointing out the change in behavior.

But happy to merge either way. going to give a few minutes for possible response then approve due to auto-merge.

mihaimaruseac: We always have the option to set this flag to false and add to exclude only a subset of these files. This is just a convenience method to ignore all at once.

spencerschrock (approved):

PR #423, opened 2025-04-04
Use `os.PathLike` instead of `pathlib.Path` in API

mihaimaruseac:
#### Summary

Public API should also work with strings. We convert to `pathlib.Path` immediately before going into the private API.

#### Release Note
NONE
#### Documentation
NONE


spencerschrock:
> Public API should also work with strings

In my experience this week, strings dont work with os.PathLike, since they dont have `__fspath__`
https://docs.python.org/3/library/os.html#os.PathLike

spencerschrock:
```
hatch run type:check
         Expected: (self, *, public_key: os.PathLike)
  Actually passed: (self, public_key: str)
  Attributes of protocol os.PathLike are not implemented on str: __fspath__

        model_signing.verifying.Config().use_elliptic_key_verifier(
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
            public_key="/tmp/foo",
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        ).set_hashing_config(
```

spencerschrock (approved):

PR #425, opened 2025-04-04
Add signature to list of paths to ignore

stefanberger:
#### Summary

Revert to the previous behavior where the signature files was automatically ignored.


mihaimaruseac (approved):
Thanks for catching this

PR #426, opened 2025-04-04
fix: argument typo and type checking in certificate Verifier

spencerschrock:
#### Summary
fix typo in X509Certificate argument name.

certifi.where returns `list[str]` but we're expecting `Iterable[pathlib.Path]` here

#### Release Note

NONE

#### Documentation

NONE


mihaimaruseac (approved):
Thank you!

Issue #179, opened 2024-05-16
Documentation clean working towards release v1


Gregory-Pereira:
**Documentation update for v1 release**

With the push towards the v1 release, documentation should be reviewed in and updated. In onboarding to the project I noticed that the model_signing docs are not current, as they utilize the upstream `google` model-signing repo instead of this one in the demo (see [code](https://github.com/sigstore/model-transparency/blob/main/model_signing/README.md?plain=1#L38)).

laurentsimon:
Another thing to do is have documentation for the API. Note: I've created a tracking issue for auto-generation / update of docs https://github.com/sigstore/model-transparency/issues/209

PR #427, opened 2025-04-04
Document the public API.

mihaimaruseac:
#### Summary
We have to expose the manifest file, but I tried to hide parts that should not be exposed externally. We don't guarantee the manifest interface to be fully stable, so this should be ok?

As a drive-by, I made the CLI code only have one single public symbol.

Part of #179.

#### Release Note
NONE
#### Documentation
NONE

### DIFF
--- a/src/model_signing/__init__.py
+++ b/src/model_signing/__init__.py
@@ -14,14 +14,114 @@
 
 """Universal model signing library.
 
-The stable high-level API is split into 3 modules:
+The API is split into 3 main components (and a glue `model_signing.manifest`
+module for data types used in public interfaces):
 
-- model_signing.hashing
-- model_signing.signing
-- model_signing.verifying
+- `model_signing.hashing`: responsible with generating a list of hashes for
+  every component of the model. A component could be a file, a file shard, a
+  tensor, etc., depending on the method used. We currently support only files
+  and file shards. The result of hashing is a manifest, a listing of hashes for
+  every object in the model.
+- `model_signing.signing`: responsible with taking the manifest and generating a
+  signature, based on a signing configuration. The signing configuration can
+  select the method used to sign as well as the parameters.
+- `model_signing.verifying`: responsible with taking a signature and verifying
+  it. If the cryptographic parts of the signature can be validated, the
+  verification layer would return an expanded manifest which can then be
+  compared agains a manifest obtained from hashing the existing model. If the
+  two manifest don't match then the model integrity was compromised and the
+  `model_signing` package detected that.
+
+The first two of these components allows configurability but can also be used
+directly, with a default configuration. The only difference is for the
+verification component where we need to configure the verification method since
+there are no sensible defaults that can be used.
+
+Signing can be done using the default configuration:
+
+```python
+model_signing.signing.sign("finbert", "finbert.sig")
+```
+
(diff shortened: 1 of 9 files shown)


src/model_signing/__init__.py
```diff
@@ -14,19 +14,119 @@
...
+).verify("finbert", "finbert.sig")
+```
+
+Alternatively, we also support automatic detection of the hashing configuration
+during the verification process. So, the following should also work:
+
+```python
+model_signing.verifying.Config().use_elliptic_key_verifier(
+    public_key="key.pub"
+).verify("finbert", "finbert.sig")
```
spencerschrock, line 99: we don't record ignore paths though? so this wouldn't be an analog of the example above?

src/model_signing/hashing.py
```diff
@@ -12,15 +12,39 @@
...
 
-Users should use this API to hash models (no signing and verification), rather
-than using the internals of the library. We guarantee backwards compatibility
-only for the API defined in `hashing.py`, `signing.py` and `verifying.py` at the
-root level of the library.
+The module could also be used to just hash a single model, without signing it:
+
+```python
+model_signing.hash(model_path)
+```
```
spencerschrock, line 24: `hashing` somewhere here?

src/model_signing/hashing.py
```diff
@@ -205,19 +238,20 @@ def use_file_serialization(
...
 
         Args:
             hashing_algorithm: The hashing algorithm to use to hash a file.
             chunk_size: The amount of file to read at once. Default is 1MB. A
               special value of 0 signals to attempt to read everything in a
               single call.
             max_workers: Maximum number of workers to use in parallel. Default
-              is to defer to the `concurrent.futures` library.
+              is to defer to the `concurrent.futures` library to select the best
+              value for the current operating system.
```
spencerschrock, line 254: nit: not operating system but machine?

src/model_signing/hashing.py
```diff
@@ -255,7 +289,8 @@ def use_shard_serialization(
               single call.
             shard_size: The size of a file shard. Default is 1 GB.
             max_workers: Maximum number of workers to use in parallel. Default
-              is to defer to the `concurrent.futures` library.
+              is to defer to the `concurrent.futures` library to select the best
+              value for the current operating system.
```
spencerschrock, line 293: same nit

src/model_signing/__init__.py
mihaimaruseac: Oh, no, that's a bug :( Let's fix that for v1.1, and I'll change the example.

src/model_signing/hashing.py
mihaimaruseac: Thanks for catching

src/model_signing/hashing.py
mihaimaruseac: Thanks, changed

src/model_signing/hashing.py
mihaimaruseac: Thanks, changed

src/model_signing/__init__.py
spencerschrock: though you're talking about the sigstore example. feel free to ignore

spencerschrock (approved):

PR #430, opened 2025-04-04
use correct module to determine ec key hash

spencerschrock:
#### Summary

`ec` refers to a module in cryptography which doesn't have a `key` symbol.

#### Release Note

NONE

#### Documentation

NONE


mihaimaruseac (approved):
Nice catch

PR #433, opened 2025-04-04
filter slsa github workflow on positive path filter only

spencerschrock:
#### Summary

My previous suggestion in 66bb7cd was wrong:
> you may only define one of `paths` and `paths-ignore` for a single
  event

This will run the workflow in cases when only a SLSA readme changes, but those are much less frequent.

#### Release Note

NONE

#### Documentation

NONE


mihaimaruseac (approved):
Nice! I saw it too just now :four_leaf_clover:

PR #439, opened 2025-04-08
cli: Use sys.exit(1) in case signing or signature verification fails

stefanberger:
#### Summary

In case signing or signature verification fails exit with '1' so that scripts can react accordingly.

### DIFF
--- a/src/model_signing/_cli.py
+++ b/src/model_signing/_cli.py
@@ -16,6 +16,7 @@
 
 from collections.abc import Iterable, Sequence
 import pathlib
+import sys
 from typing import Optional
 
 import click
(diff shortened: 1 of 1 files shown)


spencerschrock (dismissed):
Nice catch!

src/model_signing/_cli.py
```diff
@@ -224,8 +225,10 @@ def _sign_sigstore(
         ).sign(model_path, signature)
     except Exception as err:
         click.echo(f"Signing failed with error: {err}", err=True)
+        sys.exit(1)
     else:
         click.echo("Signing succeeded")
+    sys.exit(0)
```
mihaimaruseac, line 231:
```suggestion
        sys.exit(1)

    click.echo("Signing succeeded")
```

mihaimaruseac (commented):
Great catch! thank you!

Just one minor readability suggestion: since `else` on a `try`/`catch` is less frequent, let's just remove it. Also `sys.exit(0)` is default, so we can remove it.

Added the suggestion in just one place, but should be replicated

PR #441, opened 2025-04-11
Also display the fingerprint of the signing certificate

stefanberger:
#### Summary

Also display the fingerprint of the signing certficate. Adjust the ordering so that it is always shown last as the one corresponding to the signature-verifying key.

### DIFF
--- a/src/model_signing/_signing/sign_certificate.py
+++ b/src/model_signing/_signing/sign_certificate.py
@@ -175,15 +175,15 @@ class Verifier(sigstore_pb.Verifier):
         max_signing_time = signing_certificate.not_valid_before_utc
         self._store.set_time(max_signing_time)
 
-        signing_certificate_ssl = _to_openssl_certificate(
-            signing_chain.certificates[0].raw_bytes, False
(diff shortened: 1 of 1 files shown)


mihaimaruseac (dismissed):
Thank you!

src/model_signing/_signing/sign_certificate.py
```diff
@@ -175,15 +175,16 @@ def _to_openssl_certificate(certificate_bytes, log_fingerprints):
...
-        )
         trust_chain_ssl = [
             _to_openssl_certificate(
                 certificate.raw_bytes, self._log_fingerprints
             )
             for certificate in signing_chain.certificates[1:]
         ]
+        signing_certificate_ssl = _to_openssl_certificate(
+            signing_chain.certificates[0].raw_bytes,
+            self._log_fingerprints
```
mihaimaruseac, line 186: I think these still fit in the same line (`hatch fmt` should fix)

mihaimaruseac (approved):
Thank you!

PR #442, opened 2025-04-16
Add compatibility layer with v0.2 signatures

mihaimaruseac:
#### Summary
This is temporary but we need to ensure that signatures that were generated using the v0.2 CLI can still be verified at v1.0 and later.

There are 3 issues that we needed to fix:

- the PAE encoding mixed `bytes` and `str` types resulting in additional escaping being added to the statement that gets signed. Verification must replicate the same mix-up to gurantee that the signed material can be verified.
- the public key extracted from the certificate during verification was hard coded to use a hash of SHA256 instead of the one that should match the key type. Since this was using `cryptography.hazmat` there was no verification. All signatures generated at v0.2 using certificates are not really valid, but we'll support verifying them as they are.
- the in-toto predicate was different: v0.2 is not compatible with any model signature extensions, is not extensible and could be manipulated to be incomplete (not catching all possible alterations of a model). Fortunately, we used different predicate types, so it is easy to catch and add the support.

Note that this support is only added for verification size. Any new signature must be generated using the v1.0 format.

#### Release Note
Need to add to changelog

#### Documentation
Need to document the current model format better.

### DIFF
--- a/src/model_signing/_signing/sign_certificate.py
+++ b/src/model_signing/_signing/sign_certificate.py
@@ -153,7 +153,7 @@ class Verifier(sigstore_pb.Verifier):
             # Here, we patch over a bug in `pae` which mixed unicode `str` and
             # `bytes`. As a result, additional escape characters were added to
             # the material that got signed over.
-            self._public_key.verify(
+            public_key.verify(
                 envelope.signatures[0].sig,
                 sigstore_pb.pae_compat(envelope.payload),
                 ec.ECDSA(ec_key.get_ec_key_hash(public_key)),
--- a/src/model_signing/_signing/sign_certificate.py
+++ b/src/model_signing/_signing/sign_certificate.py
@@ -19,6 +19,7 @@ import logging
 import pathlib
 
 import certifi
+from cryptography import exceptions
 from cryptography import x509
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives import serialization
@@ -141,11 +142,22 @@ class Verifier(sigstore_pb.Verifier):
     def _verify_bundle(self, bundle: bundle_pb.Bundle) -> tuple[str, bytes]:
         public_key = self._verify_certificates(bundle.verification_material)
         envelope = bundle.dsse_envelope
-        public_key.verify(
-            envelope.signatures[0].sig,
-            sigstore_pb.pae(envelope.payload),
-            ec.ECDSA(ec_key.get_ec_key_hash(public_key)),
-        )
+        try:
+            public_key.verify(
+                envelope.signatures[0].sig,
+                sigstore_pb.pae(envelope.payload),
+                ec.ECDSA(ec_key.get_ec_key_hash(public_key)),
+            )
+        except exceptions.InvalidSignature:
+            # Compatibility layer with pre 1.0 release
+            # Here, we patch over a bug in `pae` which mixed unicode `str` and
+            # `bytes`. As a result, additional escape characters were added to
+            # the material that got signed over.
+            self._public_key.verify(
+                envelope.signatures[0].sig,
+                sigstore_pb.pae_compat(envelope.payload),
+                ec.ECDSA(ec_key.get_ec_key_hash(public_key)),
+            )
 
         return envelope.payload_type, envelope.payload
 
--- a/src/model_signing/_signing/sign_certificate.py
+++ b/src/model_signing/_signing/sign_certificate.py
@@ -156,7 +156,13 @@ class Verifier(sigstore_pb.Verifier):
             public_key.verify(
                 envelope.signatures[0].sig,
                 sigstore_pb.pae_compat(envelope.payload),
-                ec.ECDSA(ec_key.get_ec_key_hash(public_key)),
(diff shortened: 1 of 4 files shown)


src/model_signing/_signing/sign_sigstore_pb.py
```diff
@@ -65,6 +65,32 @@ def pae(raw_payload: bytes) -> bytes:
...
+        The encoded statement from the payload.
+    """
+    payload_type = signing._IN_TOTO_JSON_PAYLOAD_TYPE
+    payload_type_length = len(payload_type)
+    payload_length = len(raw_payload)
+    # Notice bug here!
+    pae_str = (
+        f"DSSEV1 {payload_type_length} {payload_type} "
+        f"{payload_length} {raw_payload}"
+    )
```
spencerschrock, line 90: Just pasting what the fix was to better highlight the bug. I know you described it above in docstring, I'm just a visual person :)
```python
pae_str = f"DSSEv1 {payload_type_length} {payload_type} {payload_length}"
return b" ".join([pae_str.encode("utf-8"), raw_payload])
```

src/model_signing/_signing/sign_certificate.py
```diff
@@ -141,11 +142,22 @@ def __init__(
...
+            public_key.verify(
+                envelope.signatures[0].sig,
+                sigstore_pb.pae(envelope.payload),
+                ec.ECDSA(ec_key.get_ec_key_hash(public_key)),
+            )
+        except exceptions.InvalidSignature:
+            # Compatibility layer with pre 1.0 release
+            # Here, we patch over a bug in `pae` which mixed unicode `str` and
+            # `bytes`. As a result, additional escape characters were added to
+            # the material that got signed over.
```
spencerschrock, line 155: do we want to retry all signature failures? Or just ones that declare the compatibility type?`_PREDICATE_TYPE_COMPAT`

spencerschrock (commented):
Still need to look at payload to manifest, but wanted to get partial comments out before my errand.

src/model_signing/_signing/sign_certificate.py
mihaimaruseac: We would need to do an inversion of control to check the predicate type. At this layer the predicate is just a sequence of bytes, we would need to parse it, check the predicate type and then verify that again. And then, we do the same checks again at the proper place.

The exceptions raised from the `except` block are the same as those raised from `try`, so I'd assume this would be safe.

src/model_signing/_signing/sign_sigstore_pb.py
mihaimaruseac: Yup, `"DSSEv1"` instead of `"DSSEV1"` and separating the `raw_payload` (bytes) from the string part.

src/model_signing/_signing/sign_certificate.py
spencerschrock: That was my other question, about the exception chaining. It seems like there would just be more context from both exceptions if both got raised.

src/model_signing/_signing/sign_certificate.py
```diff
@@ -141,11 +142,22 @@ def __init__(
...
+            self._public_key.verify(
+                envelope.signatures[0].sig,
```
spencerschrock, line 157: is this a typo?

`self._public_key` is used here, vs `public_key` above

src/model_signing/_signing/signing.py
```diff
@@ -120,6 +126,46 @@ def dsse_payload_to_manifest(dsse_payload: dict[str, Any]) -> manifest.Manifest:
...
+
+    Raises:
+        ValueError: The payload cannot be deserialized to a manifest.
+    """
+    # Model name is not defined, use a constant.
+    model_name = "compat-undefined-not-present"
+
+    # Serialization format is not present, build a fake one.
+    serialization = manifest.SerializationType.from_args(
+        {"method": "files", "hash_type": "sha256", "allow_symlinks": "false"}
```
spencerschrock, line 153: these fakes match the defaults from the old signatures? Though `__eq__` doesnt care about that, so I guess it doesn't matter.

spencerschrock (commented):
Looks good.

Let's make sure this works with Daniel?

src/model_signing/_signing/sign_certificate.py
mihaimaruseac: This is indeed a copy-paste typo. Fixed, thanks for catching!

src/model_signing/_signing/sign_certificate.py
mihaimaruseac: It's the same exception though. The only additional information is that this would be from a different line number.

src/model_signing/_signing/signing.py
mihaimaruseac: Yes, they should match the original from the signature. Without these fakes, we'd need to call the "private" constructor of the serialization type, but this is cleaner.

spencerschrock (approved):
> Note another bug here: the v0.2 signatures were generated with hardcoded SHA256 hash

I'll note test coverage of certificate / key signing/verifying is still pretty low at 30-40% . So maybe there's more to find :)

Issue #451, opened 2025-05-02
Signing and verifying from model directory not working (when using model_path '.')


stefanberger:
**Description**

When inside the model directory, the following command line does not work for any signing or verification method due to the internal name of the model being an empty string:

```
python -m model_signing verify sigstore \\
   [...] \\
   --signature model.sig \\
   .
```

In this case one would have to use the following work-around:  `... --signature "${PWD}"/model.sig "${PWD}"`

Two possible solutions:
- fix the paths when using '.' using `os.getcwd()`
- error out early on this instead of failing signing due to weird 'malformed in-toto statement' message; also error-out early on signature verification where not much further explanation is given

**Version**

I am using the latest upstream version.

imolloy:
I was able to do `--signature model.sig "${PWD}"`, but I had to fully-qualify paths to ignore, e.g., `--signature model.sig --ignore-paths "$PWD/.git" "${PWD}"`

PR #454, opened 2025-05-02
hashing: Fix the paths to the git related files

stefanberger:
Commit 8aba2adada812 made changes so that git-related files are not ignored anymore. Fix this by fixing their paths.

Resolves: https://github.com/sigstore/model-transparency/issues/453

##### Checklist


mihaimaruseac (commented):
This is weird, the paths should be considered relative to the model directory during serialization https://github.com/sigstore/model-transparency/blob/5c36a41fd5085b7b3949ff8ce47ba237aaa40742/src/model_signing/_serialization/serialization.py#L58-L70

stefanberger:
> This is weird, the paths should be considered relative to the model directory during serialization
> 

With this patch applied to #455 , it shows that .gitignore is now part of the signature even though it should not be.

```
diff --git a/scripts/tests/test-sign-verify b/scripts/tests/test-sign-verify
index aae7e86..6f21592 100755
--- a/scripts/tests/test-sign-verify
+++ b/scripts/tests/test-sign-verify
@@ -9,6 +9,7 @@ sigfile="${TMPDIR}/model.sig"
 echo "signme-1" > "${signfile1}"
 echo "signme-2" > "${signfile2}"
 echo "ignore" > "${ignorefile}"
+touch "${TMPDIR}/.gitignore"

 cleanup()
 {
```

#455 should be checked-in before this one. I will modify/extend test cases then to test for this.

spencerschrock:
> This is weird, the paths should be considered relative to the model directory during serialization

I believe relative paths are implied to be under the current working directory, which may differ from the model path.

https://github.com/sigstore/model-transparency/blob/b4f83c1230fd3f3428bd6f24d84ed4c478c19491/src/model_signing/hashing.py#L147-L149

Previously we did `model_path.glob` to define these paths, so it was rooted appropriately.

I think we'll have similar issues here too depending on if the user is providing relative or absolute paths.
https://github.com/sigstore/model-transparency/blob/b4f83c1230fd3f3428bd6f24d84ed4c478c19491/src/model_signing/hashing.py#L330

stefanberger:
@spencerschrock: related is #451

spencerschrock:
Yep, the new comment in that issue seems to reflect that
https://github.com/sigstore/model-transparency/issues/451#issuecomment-2856838832

spencerschrock (approved):
LGTM, only looking at b5d1ff0123285e3337a933e6fba2ed9e13398dd2 since this is based on top of #455 .

Though the test could be a normal pytest, doesn't need to be a CLI one.

spencerschrock:
Oops, forgot about the other path creation I highlighted, which would benefit from a CLI test (https://github.com/sigstore/model-transparency/issues/451#issuecomment-2856838832)
https://github.com/sigstore/model-transparency/blob/b4f83c1230fd3f3428bd6f24d84ed4c478c19491/src/model_signing/hashing.py#L330

stefanberger:
This has been resolved via #461 .

mihaimaruseac:
Oh, I missed the duplication. I think there's nothing else left in this PR?

We still need to do the same changes for the sharded hasher if I recall.

spencerschrock:
> I think there's nothing else left in this PR?

There should be in my opinion though, we ingest ignore paths in multiple places

mihaimaruseac:
Thanks for catching!

stefanberger:
> We still need to do the same changes for the sharded hasher if I recall.

Correct, we need test cases for the sharded hasher. I haven't used it at all so far.

Also, due to issues with paths I would say that for all signing tests we should verify that
- the name of the model is as-expected (the 'name' field in the subject of the in-toto statement)
- the list of signed files is as-expected
- the list of recorded ignored files is as-expected

mihaimaruseac:
Yeah, that's the purpose of the golden tests. The unit tests are more designed in the spirit of [arrange-act-assert pattern](https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/) where we might not be able to test everything in just one test.

PR #457, opened 2025-05-05
signing: Fix handling of certificate that has no KeyUsage

stefanberger:
#### Summary

If a certificate without KeyUsage is encountered then the following error is displayed:

```
Verification failed with error: \
   No <class 'cryptography.x509.extensions.KeyUsage'> extension was found
```

The error may be mistaken for a broken python cryptography library even though it is related to the certificate missing 'KeyUsage'. Fix this error by intercepting the x509.ExtensionNotFound exception and trying with the ExtendedKeyUsage next.

##### Checklist

### DIFF
--- a/src/model_signing/_signing/sign_certificate.py
+++ b/src/model_signing/_signing/sign_certificate.py
@@ -222,10 +222,14 @@ class Verifier(sigstore_pb.Verifier):
 
         extensions = signing_certificate.extensions
         can_use_for_signing = False
-        usage = extensions.get_extension_for_class(x509.KeyUsage)
-        if usage.value.digital_signature:
-            can_use_for_signing = True
-        else:
+        try:
+            usage = extensions.get_extension_for_class(x509.KeyUsage)
+            if usage.value.digital_signature:
+                can_use_for_signing = True
+        except x509.ExtensionNotFound:
+            logger.warn("Certificate does not specify 'KeyUsage'.")
+
+        if not can_use_for_signing:
             try:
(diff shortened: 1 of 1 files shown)


src/model_signing/_signing/sign_certificate.py
```diff
@@ -222,8 +222,12 @@ def _to_openssl_certificate(certificate_bytes, log_fingerprints):
 
         extensions = signing_certificate.extensions
         can_use_for_signing = False
-        usage = extensions.get_extension_for_class(x509.KeyUsage)
-        if usage.value.digital_signature:
+        try:
+            usage = extensions.get_extension_for_class(x509.KeyUsage)
+        except x509.ExtensionNotFound:
+            usage = None
```
stefanberger, line 228: Alternatively we could throw an error here not accepting certs without KeyUsage.

mihaimaruseac: I think at least logging a warning here would be good. Similarly, for the `pass` on line 240, logging would also be good there.

Then, rather than having `usage` be both a valid instance and `None`, maybe we can move the `if usage.value.digital_signature` inside the try block and change the `else` on 232 to be `if not can_use_for_signing`.

And now I realize that looking for `x509.KeyUsage` and `x509.ExtendedKeyUsage` is almost similar.

mihaimaruseac (commented):
Thank you! This looks good, I'd just suggest some logging and changing some of the logic to avoid a use of `None`

mihaimaruseac (approved):
Thank you!

Issue #429, opened 2025-04-04
hashing configuration detection can't handle ignore paths


spencerschrock:
we don't record ignore paths though? so this wouldn't be an analog of the example above?

_Originally posted by @spencerschrock in https://github.com/sigstore/model-transparency/pull/427#discussion_r2029019355_

There are probably some downsides for this either way. We'd need to think about it once we have more time after the v1 release.

spencerschrock:
In a scenario where the ignored paths belong to the model, I can see the convenience in adding them to the manifest so it can be replicated.

I am curious about what happens if a user has other files they added and want to ignore. Are there any privacy concerns by recording these files?

PR #455, opened 2025-05-03
ci: Add CLI tests and cover signatures created by v0.3.1

stefanberger:
#### Summary

Add CLI tests that can be run from the command line using `./scripts/tests/testrunner` and run it as part of the github actions.

##### Checklist

- [X] All commits are signed-off, using [DCO](https://en.wikipedia.org/wiki/Developer_Certificate_of_Origin)
- [ ] All new code has docstrings and type annotations
- [X] All new code is covered by tests. Aim for at least 90% coverage. CI is configured to highlight lines not covered by tests.
- [ ] Public facing changes are paired with documentation changes
- [ ] Release note has been added to CHANGELOG.md if needed

PR #462, opened 2025-05-14
Record and apply ignored paths

stefanberger:
#### Summary

The following changes record the signer-ignored files in the predicate of the signature and retrieve them from there when verifying. With this change it should not be necessary for a verifier to use the `--ignore-paths` option IF no changes were made to the original files and directories.

Resolves: #429 

Tests will be added once #455 has been merged.

##### Checklist

### DIFF
--- a/src/model_signing/_serialization/file.py
+++ b/src/model_signing/_serialization/file.py
@@ -17,6 +17,7 @@
 from collections.abc import Callable, Iterable
 import concurrent.futures
 import itertools
+import os
 import pathlib
 from typing import Optional
 
@@ -40,6 +41,7 @@ class Serializer(serialization.Serializer):
         *,
         max_workers: Optional[int] = None,
         allow_symlinks: bool = False,
+        ignore_paths: Iterable[pathlib.Path] = frozenset(),
     ):
         """Initializes an instance to serialize a model with this serializer.
 
@@ -55,12 +57,13 @@ class Serializer(serialization.Serializer):
         self._hasher_factory = file_hasher_factory
         self._max_workers = max_workers
         self._allow_symlinks = allow_symlinks
+        self._ignore_paths = ignore_paths
 
         # Precompute some private values only once by using a mock file hasher.
         # None of the arguments used to build the hasher are used.
         hasher = file_hasher_factory(pathlib.Path())
         self._serialization_description = manifest._FileSerialization(
-            hasher.digest_name, self._allow_symlinks
+            hasher.digest_name, self._allow_symlinks, self._ignore_paths
         )
 
     @override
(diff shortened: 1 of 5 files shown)


src/model_signing/_serialization/file.py
```diff
@@ -110,8 +111,15 @@ def serialize(
...
+        rel_ignore_paths = [
+            pathlib.Path(os.path.relpath(p, model_path)) for p in ignore_paths
+        ]
+
         return manifest.Manifest(
-            model_path.name, manifest_items, self._serialization_description
+            model_path.name,
+            manifest_items,
+            self._serialization_description,
+            rel_ignore_paths,
```
spencerschrock, line 122: open question: should we consider ignore paths as part of the serialization description, since they affect how the model was serialized?

src/model_signing/_signing/signing.py
```diff
@@ -211,6 +212,13 @@ class Payload:
       ],
       "predicateType": "https://model_signing/signature/v1.0",
       "predicate": {
+        "ignore_paths": [
+          "model.sig",
```
spencerschrock, line 216: I know this is a docstring, but would it include the signature file in the ignored paths? right now our verifier implicitly ignores the passed sig file, so I don't think it needs it again.

src/model_signing/hashing.py
```diff
@@ -140,14 +140,37 @@ def __init__(self):
         self._ignore_git_paths = True
         self.use_file_serialization()
 
-    def hash(self, model_path: PathLike) -> manifest.Manifest:
-        """Hashes a model using the current configuration."""
+    def hash(
+        self,
+        model_path: PathLike,
+        add_ignored_paths: Iterable[pathlib.Path] = frozenset(),
+    ) -> manifest.Manifest:
```
spencerschrock, line 147: this `hashing.Config` class already has a `set_ignored_paths` method.

src/model_signing/verifying.py
```diff
@@ -97,7 +97,9 @@ def verify(
 
         if self._hashing_config is None:
             self._guess_hashing_config(expected_manifest)
-        actual_manifest = self._hashing_config.hash(model_path)
+        actual_manifest = self._hashing_config.hash(
+            model_path, expected_manifest.ignore_paths
+        )
```
spencerschrock, line 102: handling manifest ignore paths could likely fall under the `if self._hashing_config is None` block instead.

spencerschrock (commented):
some initial thoughts to kick off discussion on this implementation. I also have some general thoughts I'll add to the issue

src/model_signing/_serialization/file.py
stefanberger: I can probably move it there as well.

src/model_signing/_signing/signing.py
stefanberger: True. I will try to filter it out at some point.

src/model_signing/hashing.py
stefanberger: I think what I would need is an option `add_ignored_paths()` that merge those from the manifest with the ones provided via command line.

src/model_signing/_signing/signing.py
stefanberger: Since it's being added early on in _cli.py and we don't have a standardized name `model.sig`, it's  not so easy to remove it later on.

src/model_signing/verifying.py
stefanberger: This part has changed now.

mihaimaruseac (approved):

PR #468, opened 2025-05-21
Fix handling paths to certs from `certifi.where()`

mihaimaruseac:
#### Summary

If `certifi.where()` returns a single string, we were creating multiple `pathlib.Path` objects for each character in the string, instead of wrapping the entire path string in `pathlib.Path`.

This is a type error bug, due to the fact that strings in Python are sometimes lists and sometimes not.

##### Checklist

### DIFF
--- a/src/model_signing/_signing/sign_certificate.py
+++ b/src/model_signing/_signing/sign_certificate.py
@@ -136,7 +136,13 @@ class Verifier(sigstore_pb.Verifier):
         self._log_fingerprints = log_fingerprints
 
         if not certificate_chain_paths:
-            certificate_chain_paths = [pathlib.Path(p) for p in certifi.where()]
+            os_certs = certifi.where()
+            if isinstance(os_certs, list):
+                certificate_chain_paths = [pathlib.Path(p) for p in os_certs]
+            elif isinstance(os_certs, str):
+                certificate_chain_paths = [pathlib.Path(os_certs)]
+            else:
+                raise ValueError("Could not determine signing certificates")
 
         certificates = x509.load_pem_x509_certificates(
             b"".join([path.read_bytes() for path in certificate_chain_paths])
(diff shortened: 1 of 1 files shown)


src/model_signing/_signing/sign_certificate.py
```diff
@@ -136,7 +136,13 @@ def __init__(
         self._log_fingerprints = log_fingerprints
 
         if not certificate_chain_paths:
-            certificate_chain_paths = [pathlib.Path(p) for p in certifi.where()]
+            os_certs = certifi.where()
+            if isinstance(os_certs, list):
+                certificate_chain_paths = [pathlib.Path(p) for p in os_certs]
+            elif isinstance(os_certs, str):
+                certificate_chain_paths = [pathlib.Path(os_certs)]
```
spencerschrock, line 143: > If certifi.where() returns a single string

FWIW, the library code / type hint says it will return a str.

> This is a type error bug, due to the fact that strings in Python are sometimes lists and sometimes not.

Not sure I follow.

spencerschrock (dismissed):

src/model_signing/_signing/sign_certificate.py
mihaimaruseac: Oh, if it's always a string then this is simpler, let me fix.

For the string is sometimes list, sometimes not:

```python
>>> isinstance("as", list)
False
>>> isinstance("as", str)                                                                                                                           
True
>>> isinstance(["as"], list)                                                                                                                            
True
>>> [x for x in "as"]
['a', 's']
```

In Haskell default string is always a list of characters, type checking as a list of characters. Bytestrings are just a bytestring, you cannot access the characters directly, the `in` expression does not give them one by one. I think the same is in other languages too.

spencerschrock: https://github.com/certifi/python-certifi/blob/3e7076557d66b598fcca462e422dc988f09a6802/certifi/core.py#L21

PR #471, opened 2025-05-22
feat(oauth): add --oauth_force_oob flag to support manual OAuth flow

### LINKED ISSUE #470: --oauth-force-oob should be configurable (currently defaults to manual flow and breaks OAuth with providers like Google)
**Summary**
Currently, the OAuth implementation forces an out-of-band (OOB) flow by default, as there is no way to disable it (--oauth-force-oob is not exposed or configurable). This means users are always required to manually copy-paste an auth code, and the redirect URI defaults to something like urn:ietf:wg:oauth:2.0:oob.

This causes issues with identity providers like Google, which have deprecated support for OOB redirect URIs for security reasons

**Impact:**

Users are unable to authenticate via modern OAuth providers like Google.

The CLI forces an outdated and less secure auth flow by default.

No ability to opt into a browser-based redirect flow, even when it’s the preferred experience.

SequeI:
**Summary**
This PR introduces the --oauth-force-oob flag to allow users to opt into the out-of-band (OOB) OAuth flow, instead of forcing it by default. The current behavior assumes OOB by default, which causes issues with modern OAuth providers (such as Google) that no longer support urn:ietf:wg:oauth:2.0:oob redirect URIs.

By default, the CLI will now use the standard browser-based redirect flow, improving compatibility and user experience. Users who still require OOB behavior (e.g., in headless environments) can explicitly pass the --oauth-force-oob flag.

This change makes OAuth authentication more flexible, secure, and modern by default.

```
model_signing sign bert-base-uncased --oauth_force_oob
```

```
model_signing sign bert-base-uncased
```

Resolves #470 

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ All versions prior to 1.0.0 are untracked.
 
 ## [Unreleased]
 
+- cli: `model_signing sign` now supports the `--oauth-force-oob` option (default: False)
+
 ## [1.0.1] - 2024-04-18
 
 ### Added
--- a/src/model_signing/_cli.py
+++ b/src/model_signing/_cli.py
@@ -214,6 +214,15 @@ def _sign() -> None:
         "from OIDC flow or from the environment."
     ),
 )
+@click.option(
+    "--oauth_force_oob",
+    is_flag=True,
+    default=False,
+    help=(
+        "Force an out-of-band OAuth flow and do not automatically start "
+        "the default web browser."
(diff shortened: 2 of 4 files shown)


SequeI:
Ok, that should be the final linting errors fixed.....

mihaimaruseac (approved):

mihaimaruseac:
Thank you! We might decide to change the default option before the 1.1 release, though. Going to run some experiments.

PR #478, opened 2025-06-05
add disk vs memory hasher experiment

spencerschrock:
#### Summary
This benchmark generates temporary files of varying sizes, and then hashes the file using a modified `io.SimpleFileHasher` to measure the time spent reading file chunks into memory, and the time spent to hash those chunks separately. The disk hasher reports the sum of both of those times, while the memory hasher only reports the time spent hashing.

Ensure this experiment is run in a low memory environment to minimize the effect of OS disk caching.

Below is an example in a high memory (96GiB) environment. I'll re-run it later in a low memory environment.
```
Size (B)    Memory (s)   Disk (s)
1073741824      0.6918     0.8911
2147483648      1.3805     1.7205
5368709120      3.4904     4.2287
10737418240     6.9835     8.2814
21474836480    19.3514    23.2077
53687091200    34.9285    40.8782
```

##### Checklist

### DIFF
--- a/benchmarks/exp_disk.py
+++ b/benchmarks/exp_disk.py
@@ -1,0 +1,130 @@
+# Copyright 2025 The Sigstore Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


mihaimaruseac:
> Apparently `hatch fmt` only runs `ruff format` if `ruff check` returns without issues (I had an unrelated, uncommitted file which had a few errors).

that's weird, in the past it used to run both

PR #484, opened 2025-06-17
bench: add workflow to build benchmark container

spencerschrock:
#### Summary
PR comments which start with `/bench` will trigger this workflow to build a container,
and then dispatch a job to Cloud Batch to run the benchmark.

The workflow is modeled after the OpenSSF Scorecard `scdiff` workflow, which
looks for comments from repository members. This requires the developer
triggering the benchmark to have their membership in the Sigstore
organization public.

This approach gains flexibility compared to a label trigger as
additional arguments can be provided after the /bench command.

##### Checklist


spencerschrock:
Generating models ourselves has a slight speed penalty compared to downloading, but not by much.

It's not an apples-to-apples comparison since we're generating + serializing multiple models in the first test:
30 minutes 58 seconds to generate and serialize 32, 48, and 128 GB models
25 minutes 4 seconds to download and serialize a 131 GB model.

.github/workflows/bench.yml
```diff
@@ -0,0 +1,100 @@
+name: model_signing benchmarks
+on:
+  issue_comment:
+    types: [created]
```
spencerschrock, line 4: Right now, this benchmark is triggered by `/bench` only. Perhaps a `tag` and `workflow_dispatch` trigger too?

PR #490, opened 2025-07-01
fix(signing): Handle Signature.keyid as empty string for optional values

### LINKED ISSUE #489: keyid being present in the signature due to incorrect datatype
In the protobuf specs:

```
@dataclass(eq=False, repr=False)
class Signature(betterproto.Message):
    sig: bytes = betterproto.bytes_field(1)
    """
    Signature itself. (In JSON, this is encoded as base64.)
     REQUIRED.
    """

    keyid: str = betterproto.string_field(2)
    """
    *Unauthenticated* hint identifying which public key was used.
     OPTIONAL.
    """
```

However, we assign `None` to keyid in both EC and PKSC11 signing resulting in

`"keyid": null` in the signature which is incorrect behaviour, it should just not be present if there was no keyid specified.

SequeI:
#### Summary
I left it as an empty string incase in the future we add the option for a user to specify a keyid. With this change, keyid correctly stops showing in the envelope under signatures and also is the correct type as shown in the specs. (All on the issue I opened)

```    
keyid: str = betterproto.string_field(2)
    """
    *Unauthenticated* hint identifying which public key was used.
     OPTIONAL.
    """
```

Resolves #489 

##### Checklist

### DIFF
--- a/src/model_signing/_signing/sign_ec_key.py
+++ b/src/model_signing/_signing/sign_ec_key.py
@@ -105,7 +105,7 @@ class Signer(sigstore_pb.Signer):
                 sigstore_pb.pae(raw_payload),
                 ec.ECDSA(get_ec_key_hash(self._private_key.public_key())),
             ),
-            keyid=None,
+            keyid="",
         )
 
         envelope = intoto_pb.Envelope(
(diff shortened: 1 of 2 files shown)


mihaimaruseac (approved):
Thank you!

PR #501, opened 2025-07-19
Ignore unsigned files

stefanberger:
#### Summary
This PR adds support for an option --ignore_unsigned_files to ignore files that are not part of the signature manifest. This allows to for example ignore files that were added after a signature was created. Existing test cases are extended to cover additional symlinks and regular files added after signature created and tests for expected passes and failures depending on whether --ignore_unsigned_files is used.

##### Checklist

- [X] Public facing changes are paired with documentation changes


stefanberger:
> My main worry is that we're pushing the concerns all the way to the serialization, whereas this ignore only belongs in verification. Let's still compute the manifest with all the files that exist on disk and only ignore entries during verification.

I thought that's what I was doing by only passing in the `files_to_hash` list from the verification side and leaving it `None` otherwise.

stefanberger:
... I mean it's most efficient to only hash those files that are being asked for rather than hashing all of them first and filtering after. Is this what you would want to filter-out unwanted files just before comparing the manifests?

mihaimaruseac (dismissed):
Approving instead of #502.

mihaimaruseac:
Will need to resolve conflicts, sorry for the delay that caused them

stefanberger:
Thanks for merging. 

FYI: https://github.com/vllm-project/vllm/pull/21957

PR #502, opened 2025-07-21
Ignore unsigned files (read and hash all files and filter late)

stefanberger:
#### Summary

This PR adds support for an option --ignore_unsigned_files to ignore files that are not part of the signature manifest. This allows to for example ignore files that were added after a signature was created. Existing test cases are extended to cover additional symlinks and regular files added after signature created and tests for expected passes and failures depending on whether --ignore_unsigned_files is used.

Compared to PR #501, the implementation in this PR still has all files read and hashed and then ensures that while comparing the files (items) in the `actual_manifest` are not a true subset of the files (items) in the `expected_manifest`. If the `actual_manifest` were a true subset of the `expected_manifest`, this would mean that the `expected_manifest` was missing files.

##### Checklist

### DIFF
--- a/scripts/tests/test-sign-verify
+++ b/scripts/tests/test-sign-verify
@@ -160,6 +160,40 @@ fi
 check_model_name "${sigfile}" "$(basename "${TMPDIR}")"
 
 echo
+echo "Creating a symlink, that is not part of the signature, to make signature verification fail (1)"
+
+echo "foo" > symlinked
+ln -s symlinked symlink
+
+if python -m model_signing \
+	verify certificate \
+	--signature "$(basename "${sigfile}")" \
+	--certificate_chain "${DIR}/keys/certificate/ca-cert.pem" \
+	--ignore-paths "$(basename "${ignorefile}")" \
+	. ; then
+	echo "Error: 'verify certificate' succeeded after new file (symlink) was created"
+	exit 1
+fi
+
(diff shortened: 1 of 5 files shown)


mihaimaruseac (approved):
This is closer to the plans to verify only the files for the platform the uses (e.g., on PyTorch, whereas HuggingFace model repo holds files for all frameworks under the same name (& signature, in future))

scripts/tests/test-sign-verify
```diff
@@ -159,6 +159,40 @@ if [ "${res}" != "${exp}" ]; then
...
+# This should pass when passing --allow_symlinks when hashing all the files
+echo
+echo "Pass signature verification by ignoring any unsigned files or symlinks"
+if ! python -m model_signing \
+	verify certificate \
+	--signature "$(basename "${sigfile}")" \
+	--certificate_chain "${DIR}/keys/certificate/ca-cert.pem" \
+	--ignore-paths "$(basename "${ignorefile}")" \
+	--ignore_unsigned_files \
+	--allow_symlinks \
```
stefanberger, line 187: That this --allow-symlinks needs to be passed here to pass the test may be 'odd'. A symlink that is supposed to be ignored has to be allowed to be hashed first and then will be ignord...

mihaimaruseac:
Closing in favor of #501, thank you for opening both and illustrating the problems with this one

PR #458, opened 2025-05-06
cli: add otel trace auto-instrumentation

miyunari:
#### Summary

This PR adds OpenTelemetry-SDK auto-instrumentation to the model-validation-cli.

##### Checklist

- [ ] All commits are signed-off, using [DCO](https://en.wikipedia.org/wiki/Developer_Certificate_of_Origin)
- [ ] All new code has docstrings and type annotations
- [ ] All new code is covered by tests. Aim for at least 90% coverage. CI is configured to highlight lines not covered by tests.
- [ ] Public facing changes are paired with documentation changes
- [ ] Release note has been added to CHANGELOG.md if needed

PR #503, opened 2025-07-22
cli: add otel trace auto-instrumentation

knrc:
#### Summary
This PR extends @miyunari's #458 to address the comments made by @mihaimaruseac.

The instructions for running this are the same as those mentioned in #458, with the exception that the opentelemetry is now an optional dependency.  This dependency can be installed through `pip install model-signing[otel]`, and will fallback to a noop tracer if the import is unsatisfied.

##### Checklist

- [X] Public facing changes are paired with documentation changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,61 @@ the PKCS #11 device and store it in a file in PEM format. With can then use:
        --public_key key.pub  /path/to/your/model
 ```
 
+#### OpenTelemetry Support
+
+Model signing supports optional distributed tracing and observability through OpenTelemetry. This allows you to monitor signing operations, track performance, and integrate with observability platforms.
+
+To enable OpenTelemetry support, install the optional dependencies:
+
+```bash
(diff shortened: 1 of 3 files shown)


knrc:
It looks like there are a couple of linting failures, I'll tidy those up shortly

knrc:
@mihaimaruseac No worries.  There's obviously something else going on with the linting as I had these running clean using `act`, I'll need to dig deeper into what is going on within the github runs.

knrc:
> Do we want to also include the otel deps in our image? Will add depending if the answer is yes whichever PR gets merged first

That wasn't in @miyunari's original PR, and since the dependencies are now optional I don't believe it's necessary for this PR.  Would we want to have two images, with and without the tracing?

mihaimaruseac:
I would say to get all dependencies included in the image?

Or maybe create 2 images? One with minimal deps and one with everything?

PR #514, opened 2025-08-05
Add support to hash via GPUs

hotcodemacha:
#### Summary

Closes https://github.com/sigstore/model-transparency/issues/201

Introduced a new PyTorch-powered TorchSHA256 hashing engine that chooses a CUDA GPU when available and otherwise falls back to CPU, implementing the complete SHA256 algorithm with explicit 32-bit masking and standard round constants

Exposed a [gpu] optional dependency group to install torch for users who want GPU hashing support

Added tests verifying that the GPU-backed engine produces the same digest and algorithm name as the existing CPU SHA256 implementation

##### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,6 +59,10 @@ otel = [
   "opentelemetry-exporter-otlp",
 ]
 
+gpu = [
+  "torch",
+]
+
 [project.scripts]
 model_signing = "model_signing._cli:main"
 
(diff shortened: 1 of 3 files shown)


mihaimaruseac (commented):
I think we actually want the hashing itself to be via a C/C++ module that runs directly on GPU, not via a deep learning library. There is already some work being done by Purdue to hash models on GPU.

ashutoshcipher:
> I think we actually want the hashing itself to be via a C/C++ module that runs directly on GPU, not via a deep learning library. There is already some work being done by Purdue to hash models on GPU.

ack

PR #520, opened 2025-08-20
Update serialization metadata when toggling symlink handling

hotcodemacha:
#### Summary

Documented the ignore_paths parameter in the file serializer and ensured symlink policy changes update the recorded serialization metadata

Updated the file-shard serializer to refresh its serialization metadata when toggling symlink handling

Added tests verifying that manifests correctly reflect updated symlink settings for both file and shard serializers

##### Checklist

### DIFF
--- a/src/model_signing/_serialization/file.py
+++ b/src/model_signing/_serialization/file.py
@@ -53,6 +53,7 @@ class Serializer(serialization.Serializer):
             allow_symlinks: Controls whether symbolic links are included. If a
               symlink is present but the flag is `False` (default) the
               serialization would raise an error.
+            ignore_paths: The paths of files to ignore.
         """
         self._hasher_factory = file_hasher_factory
         self._max_workers = max_workers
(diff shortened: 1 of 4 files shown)


mihaimaruseac (changes requested):
I thought this is already handled in `serialize`? We always need to call `serialize` and that already captures the correct information.

Was there a bug you identified?

hotcodemacha:
> I thought this is already handled in `serialize`? We always need to call `serialize` and that already captures the correct information.
> 
> Was there a bug you identified?

The bug was in the internal metadata `(_serialization_description)` that’s embedded in each manifest.
`Serializer.__init__` precomputes that metadata using the initial `allow_symlinks` value.
Calling `set_allow_symlinks()` updated `self._allow_symlinks`, but the precomputed metadata stayed stale; the next call to `serialize()` reused the old metadata and wrote the wrong symlink setting into the manifest.
The fix recomputes _serialization_description when `set_allow_symlinks()` is called so that every subsequent `serialize()` emits a manifest reflecting the current symlink policy.

PR #533, opened 2025-09-24
Replace sigstore_proto_buf with sigstore_models

stefanberger:
#### Summary

Replace sigstore_proto_buf with sigstore_models. In many cases the classes of sigstore_modelscan be called with unchanged parameters, but in some cases explicit base64 encoding needs to be done.

##### Checklist

### DIFF
--- a/src/model_signing/_signing/sign_certificate.py
+++ b/src/model_signing/_signing/sign_certificate.py
@@ -14,6 +14,7 @@
 
 """Signers and verifiers using certificates."""
 
+import base64
 from collections.abc import Iterable
 import logging
 import pathlib
@@ -26,8 +27,8 @@ from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import ec
 from cryptography.x509 import oid
 from OpenSSL import crypto
-from sigstore_protobuf_specs.dev.sigstore.bundle import v1 as bundle_pb
-from sigstore_protobuf_specs.dev.sigstore.common import v1 as common_pb
+from sigstore_models.bundle import v1 as bundle_pb
+from sigstore_models.common import v1 as common_pb
 from typing_extensions import override
 
 from model_signing._signing import sign_ec_key as ec_key
@@ -79,8 +80,10 @@ class Signer(ec_key.Signer):
     def _get_verification_material(self) -> bundle_pb.VerificationMaterial:
         def _to_protobuf_certificate(certificate):
             return common_pb.X509Certificate(
-                raw_bytes=certificate.public_bytes(
-                    encoding=serialization.Encoding.DER
+                raw_bytes=base64.b64encode(
+                    certificate.public_bytes(
+                        encoding=serialization.Encoding.DER
+                    )
                 )
             )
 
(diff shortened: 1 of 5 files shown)


src/model_signing/_signing/sign_certificate.py
```diff
@@ -95,7 +98,8 @@ def _to_protobuf_certificate(certificate):
         return bundle_pb.VerificationMaterial(
             x509_certificate_chain=common_pb.X509CertificateChain(
                 certificates=chain
-            )
+            ),
+            tlog_entries=[],
```
stefanberger, line 102: @woodruffw Should this have to be here or is this a bug in the library?

mihaimaruseac:
> I got a lot farther now but had to disable verification tests with public keys because sigstore-models does seem to deal with plain public keys the same way as protobuf did.

Unfortunately this would break verifying models from the NVidia hub in that case, so we'd still need to patch to support that use case.

stefanberger:
> > I got a lot farther now but had to disable verification tests with public keys because sigstore-models does seem to deal with plain public keys the same way as protobuf did.
> 
> Unfortunately this would break verifying models from the NVidia hub in that case, so we'd still need to patch to support that use case.

We do have test cases verifying older signatures and they do pass now...
Also, do you have a link to a model on the Nvidia hub? I seem to not have looked at the right place to find something to git clone and verify.

stefanberger:
> One sample model: https://catalog.ngc.nvidia.com/orgs/nvidia/teams/tao/models/peoplenet?version=pruned_quantized_decrypted_v2.3.4
> 
> There are also intructions on https://developer.nvidia.com/blog/bringing-verifiable-trust-to-ai-models-model-signing-in-ngc/

Whatever 'File Browser' I clicked on of various versions, I don't see any file that looks like model.sig. 

https://catalog.ngc.nvidia.com/orgs/nvidia/teams/tao/models/peoplenet/files?version=pruned_quantized_decrypted_v2.3.4

stefanberger:
> Yeah, the signature is not displayed there, you download it with the CLI command

I must be doing something wrong. With the command `ngc registry model download-version "nvidia/tao/peoplenet:pruned_quantized_decrypted_v2.3.4"` copied to my clipboard from the (assumed) signed model at https://catalog.ngc.nvidia.com/orgs/nvidia/teams/tao/models/peoplenet/files?version=pruned_quantized_decrypted_v2.3.4 I get exactly the files that it shows on the webpage and none of them seems to be a signature:
```
$ ls -l
total 8612
-rw-------. 1 stefanb stefanb      16 Sep 29 08:16 labels.txt
-rw-------. 1 stefanb stefanb     207 Sep 29 08:16 nvinfer_config.txt
-rw-------. 1 stefanb stefanb 8788899 Sep 29 08:16 resnet34_peoplenet_int8.onnx
-rw-------. 1 stefanb stefanb   16678 Sep 29 08:16 resnet34_peoplenet_int8.txt
```

mihaimaruseac:
Found the instructions on how to get the signature: see step 2 at https://developer.nvidia.com/blog/bringing-verifiable-trust-to-ai-models-model-signing-in-ngc/

PR #537, opened 2025-10-07
Add more tests

stefanberger:
#### Summary
Add test cases for v1.0.1 signatures.
Also add a test that first creates key and certificate type of signatures with the currently active library and tests verification against several older versions of the library that are installed in a venv. This serves the purpose to ensure backwards compatibility. Then also use the older version of the library and model_signing tool to verify against the key, certificate, and sigstore signatures we have in the test directory.

##### Checklist


stefanberger:
Putting this into draft mode while I experiment with the oidc-token from [here](https://raw.githubusercontent.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon/refs/heads/current-token/oidc-token.txt) for sigstore signing.

stefanberger:
It can now sign with sigstore using that test-token as well. It has problems with --use_staging, though, so I cannot use this option.

spencerschrock:
> It has problems with --use_staging, though, so I cannot use this option.

This is specifically for trying to verify new signatures on old versions? I believe rekor v2 is live in staging, which is why you're having this issue (verifying a rekorv2 entry with sigstore-python v3.6.5 which doesnt have support for them yet?) .

I understand verifying old signatures with new versions, but do we need tests for verifying new signatures using old versions?

Issue #160, opened 2024-04-17
API for incremental model re-computation


laurentsimon:
Provide an API that allows re-computing a subset of files in a model. This is useful in cases where only a (small) set of files have changes, eg if a user has updated a README in a huggingface repo.

The API may look something like the following:

```
sign(..., updated_files)
```

with 

```
updated_files = list [
   (src_file, action, dst_file)
]
```

Example:

```
updated_files = [
  "path/to/file1", "RENAMED", "path/to/file2",
 "path/to/file3", "DELETED", "",
  "path/to/file4", "UPDATED", "path/to/file4"
]
```

Something along these lines.  The API would use the existing manifest file and update it by re-computing only files that need to be re-hashed.

laurentsimon:
A simple first iteration could be to support just a list of files to recompute a hash for. The serializer would list files on disk, and for those that needs to be recomputed, it would re-serialize them. For the others, it would take it from the existing signature file.

mihaimaruseac:
I think we always need 2 lists: one for files that got deleted and one for files that need to be rehashed.

For a deleted file, we just remove it from the manifest, error if it was not present.

For files that need to be rehashed, error if they're not present, recompute hash otherwise.

Moving/renaming files and just copying the hash will speed up the update process as we won't need to re-read the file's content, but we can risk a misuse of the API where a file that's both moved and modified will have the wrong hash.

laurentsimon:
> I think we always need 2 lists: one for files that got deleted and one for files that need to be rehashed.

> For a deleted file, we just remove it from the manifest, error if it was not present.
> 
> For files that need to be rehashed, error if they're not present, recompute hash otherwise.

Depends what the semantics we define. If we allow files in the list to not be present on disk,  it would allow the caller to pass a list of changed files (either deleted, renamed or added) without the need for them to know the details of what was changed. Our lib would figure what to recompute by listing existing files on disk and using the list to determine which to recompute (deleted files would simply be ignored since they won't be on disk).

I'm thinking that's as simple as it gets for callers. They would be able to use existing git command to list files changed, without parsing what sort of changes were made. To be honest I have not not played with git yet to see what command is needed to get the diff between a commit and the latest push :)

A slightly more advanced solution is for callers to pass the type of changes (either via 2 lists as you suggest) or via a more structured parameters (in the original description).

mihaimaruseac:
[...]

PR #562, opened 2025-11-02
feat: Add incremental re-hashing API for large models

edonadei:
## Motivation

This PR implements incremental model re-hashing to solve a critical performance problem when signing large ML models. Currently, when a user makes a small change to a large model (e.g., updating README.md in a 500GB model), the entire model must be re-hashed before re-signing, which can take hours. This makes it impractical to update documentation or configuration files in large models.

This PR adds a Python API that reuses digests from previous signatures for unchanged files, only re-hashing files that were added or modified. For a 500GB model with a 1KB documentation update, this reduces re-hashing time from hours to seconds (~500,000x speedup).

## Changes

1. **`signing.manifest_from_signature()`** - Extracts and **verifies** manifest from existing signature files, ensuring old signatures haven't been tampered with before reusing their hashes
2. **`IncrementalSerializer`** - Core implementation that compares current model state against existing manifest and only re-hashes changed files
3. **`Config.use_incremental_serialization()`** - Integrates incremental serializer into hashing API
4. **`sign_incremental()`** - High-level convenience API that combines verification + extraction + incremental hashing + signing

## Design Decision

Following the discussion in #160, this implementation uses a user-driven approach where changed files are specified via the files_to_hash parameter (e.g., from git diff). This was chosen over automatic change detection because:
- File metadata (mtime) is unreliable across systems and git operations
- Users know what changed via their workflow (git, manual tracking, etc.)
- Keeps implementation simple and reliable
- Works with any file tracking system

The implementation also reuses the existing `Verifier.verify()` path instead of manually parsing DSSE envelopes, which:
- Eliminates code duplication
- Ensures consistent verification logic
- Adds security by verifying old signatures before trusting their hashes

## Test Coverage:
- tests/manifest_test.py: Tests for manifest data structures
- tests/_serialization/incremental_test.py: 7 tests for incremental serializer (new files, modified files, deleted files, mixed scenarios, empty manifests)
- tests/hashing_config_test.py: Tests for incremental serialization parameter extraction
- tests/_signing/signing_test.py: Tests for DSSE payload parsing and verification

## Future Work: Based on maintainer feedback, these could be added in follow-up PRs:
- CLI support (--incremental flag)
- Documentation in README
- Git helper utility (get_changed_files_from_git())
- Integration tests with large models

## Questions for Maintainers:
- Is the files_to_hash parameter approach acceptable, or would you prefer a different change detection mechanism? ✅  Looks good
- Should CLI support and documentation be in this PR or separate follow-ups? ✅ Will we done in a next PR
- Is the breaking change (adding required verification parameters) acceptable for the incremental signing API?

## Testing this PR

```python
# Create and sign an initial model
from model_signing import signing
from pathlib import Path

model_dir = Path("test_model")
model_dir.mkdir()
(model_dir / "weights.bin").write_bytes(b"x" * 1000000)  # 1MB file
(model_dir / "README.md").write_text("Version 1")

# Initial sign
signing.Config().use_elliptic_key_signer(
    private_key=Path("test.key")
).sign(model_dir, "model.sig.v1")

# Modify only README
(model_dir / "README.md").write_text("Version 2 - updated docs")

# Sign incrementally - only re-hashes README.md, reuses weights.bin digest
# Must verify the old signature before reusing its hashes
signing.Config().use_elliptic_key_signer(
    private_key=Path("test.key")
).sign_incremental(
    model_dir,
    old_signature_path="model.sig.v1",
    new_signature_path="model.sig.v2",
    identity="test-identity@example.com",
    oidc_issuer="https://example.com/oauth",
    files_to_hash=[model_dir / "README.md"]
)

# Verify the new signature works
from model_signing import verifying
verifying.Config().use_elliptic_key_verifier(
    public_key=Path("test.pub")
).verify(model_dir, "model.sig.v2")
```

### DIFF
--- a/src/model_signing/_serialization/incremental.py
+++ b/src/model_signing/_serialization/incremental.py
@@ -25,7 +25,6 @@ import itertools
 import logging
 import os
 import pathlib
-from typing import Optional
 
 from typing_extensions import override
 
@@ -57,7 +56,7 @@ class IncrementalSerializer(serialization.Serializer):
         file_hasher_factory: Callable[[pathlib.Path], io.FileHasher],
         existing_manifest: manifest.Manifest,
         *,
-        max_workers: Optional[int] = None,
+        max_workers: int | None = None,
         allow_symlinks: bool = False,
         ignore_paths: Iterable[pathlib.Path] = frozenset(),
     ):
@@ -95,12 +94,26 @@ class IncrementalSerializer(serialization.Serializer):
                 if isinstance(item, manifest._File):
                     self._existing_items[item.path] = item
 
-        # Precompute serialization description
-        hasher = file_hasher_factory(pathlib.Path())
-        self._serialization_description = manifest._FileSerialization(
-            hasher.digest_name, self._allow_symlinks, self._ignore_paths
-        )
-        self._is_blake3 = hasher.digest_name == "blake3"
+        # Copy serialization description from existing manifest
+        if isinstance(
+            existing_manifest._serialization_type, manifest._FileSerialization
+        ):
+            # Copy from file-based serialization
+            self._serialization_description = manifest._FileSerialization(
+                existing_manifest._serialization_type._hash_type,
+                existing_manifest._serialization_type._allow_symlinks,
+                existing_manifest._serialization_type._ignore_paths,
+            )
+            self._is_blake3 = (
+                existing_manifest._serialization_type._hash_type == "blake3"
+            )
+        else:
+            # Fall back to creating new one for shard-based manifests
+            hasher = file_hasher_factory(pathlib.Path())
+            self._serialization_description = manifest._FileSerialization(
+                hasher.digest_name, self._allow_symlinks, self._ignore_paths
+            )
+            self._is_blake3 = hasher.digest_name == "blake3"
 
     def set_allow_symlinks(self, allow_symlinks: bool) -> None:
         """Set whether following symlinks is allowed."""
@@ -177,7 +190,7 @@ class IncrementalSerializer(serialization.Serializer):
         model_path: pathlib.Path,
         *,
         ignore_paths: Iterable[pathlib.Path] = frozenset(),
-        files_to_hash: Optional[Iterable[pathlib.Path]] = None,
+        files_to_hash: Iterable[pathlib.Path] | None = None,
     ) -> manifest.Manifest:
         """Serializes the model, only re-hashing changed/new files.
 
--- a/src/model_signing/hashing.py
+++ b/src/model_signing/hashing.py
@@ -380,13 +380,11 @@ class Config:
         self,
         existing_manifest: manifest.Manifest,
         *,
-        hashing_algorithm: Optional[
-            Literal["sha256", "blake2", "blake3"]
-        ] = None,
+        hashing_algorithm: Literal["sha256", "blake2", "blake3"] | None = None,
         chunk_size: int = 1048576,
-        max_workers: Optional[int] = None,
-        allow_symlinks: Optional[bool] = None,
-        ignore_paths: Optional[Iterable[pathlib.Path]] = None,
+        max_workers: int | None = None,
+        allow_symlinks: bool | None = None,
+        ignore_paths: Iterable[pathlib.Path] | None = None,
     ) -> Self:
         """Configures incremental serialization for selective file re-hashing.
 
--- a/src/model_signing/signing.py
+++ b/src/model_signing/signing.py
@@ -83,7 +83,7 @@ def sign_incremental(
     identity: str,
     oidc_issuer: str,
     use_staging: bool = False,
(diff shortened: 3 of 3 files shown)


src/model_signing/_serialization/incremental.py
```diff
@@ -0,0 +1,247 @@
...
+                # File was explicitly marked as changed - must re-hash it
+                needs_rehash = True
+            elif not rehash_paths:
+                # No explicit files_to_hash provided, so we're in "scan mode"
+                # Reuse digest for existing files (assume unchanged)
+                needs_rehash = False
+            else:
+                # File exists in old manifest and wasn't marked as changed
+                # Reuse old digest
+                needs_rehash = False
```
mihaimaruseac: This block should be extracted to own utility function (since it will likely be reused when adding support for incremental hashing of shards). It also improves readability, as this function right now is quite large.

src/model_signing/_serialization/incremental.py
```diff
@@ -0,0 +1,247 @@
...
+        # Build lookup dictionary: file path -> manifest item
+        self._existing_items = {}
+        for item in existing_manifest._item_to_digest:
+            # item is a _File or _Shard key; we only support files for now
+            if isinstance(item, manifest._File):
+                self._existing_items[item.path] = item
+
+        # Precompute serialization description
+        hasher = file_hasher_factory(pathlib.Path())
+        self._serialization_description = manifest._FileSerialization(
```
mihaimaruseac: This would be the same as `manifest.serialization_type()`, right?

src/model_signing/manifest.py
```diff
@@ -466,3 +468,51 @@ def serialization_type(self) -> dict[str, Any]:
         manifest so that signature verification can use the same method.
         """
         return self._serialization_type.serialization_parameters
+
+    @classmethod
+    def from_signature(cls, signature_path: pathlib.Path) -> Self:
```
mihaimaruseac: This should probably call the verification path first and then use https://github.com/sigstore/model-transparency/blob/992a56ba2c4f84727729498e5dbd2c16ad5b0d7c/src/model_signing/_signing/signing.py#L353 to get the manifest directly without duplication.

mihaimaruseac (changes requested):
I like the general direction of this, but let's try to reuse existing infrastructure as much as possible, instead of duplicating.

I didn't look at the tests too much, but I noticed we have new hardcoded values. Can we use the existing goldens approach for this? Or use the fixture mechanism?

mihaimaruseac:
> * Is the files_to_hash parameter approach acceptable, or would you prefer a different change detection mechanism?

This is perfect. We'll have users specify these in advance, in a clean pipeline, rather than us having to write and maintain the detection logic

> * Should CLI support and documentation be in this PR or separate follow-ups?

It's ok to do it after.

> * Would you like me to add a CHANGELOG entry now or wait until the approach is approved?

Same here for later, assuming we collect everything

src/model_signing/_serialization/incremental.py
```diff
@@ -0,0 +1,247 @@
...
+            hasher.digest_name, self._allow_symlinks, self._ignore_paths
+        )
+        self._is_blake3 = hasher.digest_name == "blake3"
+
+    def set_allow_symlinks(self, allow_symlinks: bool) -> None:
```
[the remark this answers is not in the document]
edonadei: Added a warning log.

src/model_signing/_serialization/incremental.py
mihaimaruseac: Oh, that's a nicer improvement!

src/model_signing/_serialization/incremental.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+        self._hasher_factory = file_hasher_factory
+        self._existing_manifest = existing_manifest
+        self._max_workers = max_workers
+        self._allow_symlinks = allow_symlinks
+        self._ignore_paths = ignore_paths
+
+        # Build lookup dictionary: file path -> manifest item
+        self._existing_items = {}
+        for item in existing_manifest._item_to_digest:
```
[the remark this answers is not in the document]
mihaimaruseac: Yeah, we can do incremental just for files now, but on the longer term, I think we should support it for the shards too

src/model_signing/manifest.py
edonadei: Ok that's what I tried to do! Will refresh the review for you to take a second look at it

src/model_signing/_serialization/incremental.py
mihaimaruseac: I think this should still just copy from the existing manifest (similar to a C++ copy-constructor)

edonadei:
Addressed your last comment @mihaimaruseac

edonadei:
Tried to re-run tests and checks locally. Hope it works now

PR #564, opened 2025-11-05
(fix): copy only relevant files onto the build stages

sampras343:
#### Summary

This PR refactors the Docker build stages so that we only copy the files required to build and install the Python package instead of copying the entire repository. Our current image build pulls in demo/example content that isn’t part of the final deliverable, and that content drags in dependencies which are flagged by scanners.

#### Changes
Narrow copy in builder stages
Instead of COPY . /app, we now do:
```
COPY src /app/src
COPY pyproject.toml /app/
COPY README.md /app/
COPY LICENSE /app/
```

in both minimal_install and full_install.

##### Why

- Fixes the root cause: the root cause was “too much gets copied into the build stage,” not “the project depends on Keras.” Narrowing the build context removes the source of the false-positive CVE.
- Keeps final image small and clean: by only copying the installed artifacts from the builder into the runtime image, we avoid bringing tutorial/example code into the final image.
- Matches packaging intent: Python projects that are meant to be installed usually only need src/ + project metadata. Copying the whole repo is extra and risky.

##### Security Context
- CVEs were detected because the build layer contained files/deps that reference Keras, not because the final image uses that vulnerable code path.
- Mitigation in this PR: remove the non-deliverable content from the build layer so scanners no longer observe that vulnerable dependency in this image.

##### Checklist

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -26,12 +26,18 @@ RUN apt-get update && \
 
 FROM base_builder AS minimal_install
 WORKDIR /app
-COPY . /app
+COPY src /app/src
+COPY pyproject.toml /app/
+COPY README.md /app/
+COPY LICENSE /app/
 RUN pip install .
 
 FROM base_builder AS full_install
 WORKDIR /app
-COPY . /app
+COPY src /app/src
+COPY pyproject.toml /app/
+COPY README.md /app/
+COPY LICENSE /app/
 RUN pip install .[pkcs11,otel]
 
 FROM base AS minimal_image


mihaimaruseac (approved):
Thank you!

spencerschrock (approved):

sampras343:
> I think this was also copying SLSA related code and we pin to older versions of Keras for now (as just one example). That would get fixed, either by removing from the repo or updating the demos to newer versions.

Yes, you are right @mihaimaruseac. And thanks for the merge

PR #566, opened 2025-11-05
Digest subcommand

puerco:
#### Summary
This PR adds a `digest` subcommand to the model_signing CLi. This allows other tools to find out the computed digest of a serialized model directory. It uses the same serializer as the sign command and supports adding files to the ignore list.

```
[puerco@babieco model-transparency] on  digest-subcommand 🐍 [venv]❯ python -m model_signing digest --help
Usage: python -m model_signing digest [OPTIONS] MODEL_PATH

  Computes the digest of a model.

  The digest subcommand serializes a model directory and computes the "root"
  digest (hash), the same used when signing and as the attestation subject.

  By default, git-related files are ignored (same behavior as the sign
  command). Use --no-ignore-git-paths to include them. To ignore other files
  from the directory serialization, use --ignore-paths.

Options:
  --ignore-paths IGNORE_PATHS     File paths to ignore when signing or
                                  verifying.
  --ignore-git-paths / --no-ignore-git-paths
                                  Ignore git-related files when signing or
                                  verifying.  [default: ignore-git-paths]
  --allow_symlinks                Whether to allow following symlinks when
                                  signing or verifying files.
  -h, --help                      Show this message and exit.

```

##### Checklist

Closes https://github.com/sigstore/model-transparency/issues/565

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ All versions prior to 1.0.0 are untracked.
 ## [Unreleased]
 
 ### Added
-- ...
+-Added the `digest` subcommand to compute and print a model's digest. This enables other tools to easily pair the attestations with a model directory.
 
 ### Changed
 - ...


ralphbean:
As a user, I see the value in getting just the root hash. I would use it for the Subject of _other attestations_.

Outputting all the hashes in the manifest would only be valuable to me in helping explain and understanding what the tool has done. The serialization field would be necessary too, if all the resources are also necessary.

mihaimaruseac:
That makes sense. What about displaying just the root hash by default, but if the user passes a `--full` flag we return a json that contains all the hashes and the serialization field?

mihaimaruseac:
Review on Jan 14: `--full` could be added later, so this is ready to merge

PR #581, opened 2025-12-04
feat: Support keys and signatures from memory

### LINKED ISSUE #559: Support more key formats
**Description**

Currently the only way to provide an elliptic key manually is to have a PEM-encoded public key written to the disk, and then pass that to `use_elliptic_key_verifier`. It would be helpful if once could also pass the 33-byte compressed version of the public key, just as bytes.

SequeI:
Add support for providing elliptic curve keys and signatures as bytes in the library API. Private/public keys can now be passed as bytes (PEM or compressed format), and signatures can be provided as JSON strings or bytes. All changes are backward compatible.

- Signer/Verifier accept keys from bytes or file paths
- Public keys support compressed format (33/49/67 bytes)
- Signature.read() and verify() accept bytes/strings
- Updated documentation and changelog

@makew0rld Please let me know if this is what you had in mind, if anything is missing etc :)

Closes #559 

##### Checklist

### DIFF
--- a/src/model_signing/_signing/sign_ec_key.py
+++ b/src/model_signing/_signing/sign_ec_key.py
@@ -21,7 +21,6 @@ from cryptography import exceptions
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import ec
-from cryptography.hazmat.primitives.asymmetric import types as crypto_types
 from google.protobuf import json_format
 from sigstore_models import intoto as intoto_pb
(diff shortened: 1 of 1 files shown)


SequeI:
> One other related thing that would be nice is the ability to get the signature as bytes/text rather than having to write it to disk.

Good idea. Maybe also a `--stdout` cli bool flag as an alternate to a defined signature path. I think it would do better in a separate PR though as the one has a good amount of API changes as is. Will open an issue to track this! Thanks

Once this is in, I'll push up the second PR

Issue #586, opened 2025-12-10
OMS model signature and other attestations


ralphbean:
The AI/ML supply chain security landscape involves multiple attestation types which can be represented as in-toto attestations. For example:

* **OpenSSF Model Signing (OMS) Signatures** help verify integrity and authenticity
* **SLSA Provenance Attestations** describe how the model was built
* **SBOMs (Software/AI Bill of Materials)** inventory the components, training datasets, foundation models, and dependencies that comprise the model

All three can be represented as in-toto statements with different [predicates](https://github.com/in-toto/attestation/tree/main/spec/predicates), wrapped in a sigstore bundle. The OMS model-signing signatures are themselves currently represented as an in-toto statement wrapped in a sigstore bundle in a single `model.sig` file, detached but typically co-located with the model files.

How should other attestations relate to the OMS signature? Critically, **should the OMS model signature include the hash of other attestation bundle files itself?**

Attestations can accumulate throughout a model's lifecycle—from training through deployment and operation. Adding new attestations (e.g., security scan results, VEX statements, post-deployment audits) after initial release is a strength we should try to preserve.

How do other attestation sigstore bundles appear in the layout of the model repository? There’s no convention or rule for how they must be named. They may appear as individual files, like `model.spdx.sigstore.json`, or they might all be appended to a growing [jsonl file](https://jsonlines.org/) named something like `claims.jsonl`. (See also [sigstore/cosign#4496](https://github.com/sigstore/cosign/issues/4496).)

If the jsonl list convention is not used, there is no problem. New attestations can be added without breaking OMS signature validation of the model.

If the `claims.jsonl` list convention is used and if the hash of that file is included in the OMS signature, then any future attestations that get appended to that file will change the hash and break the integrity check of the model as a whole.

Is this desirable? A downstream consumer may want to detect that an originally present SLSA provenance attestation that was present in the model at the time of signing is no longer present. However, if a consumer cares about that particular attestation type and its signing identity, they can use a policy mechanism to require its presence, regardless of the OMS signature integrity.

**Position**: The model signature should NOT include the bundle file hashes of other attestations.

**Position**: We will NOT expand the OMS predicate to include information that can be better represented in other orthogonal attestations.

[...]

PR #590, opened 2025-12-17
Add configurable exclusion of attestation files from model signatures

ralphbean:
Adds support for excluding attestation files (signature and attestation bundle files) from model signatures by default, addressing #586.

## Changes

- Attestation files (`*.sig`, `*.sigstore.json`, `claims.jsonl`) are now excluded from model signatures by default
- New `--ignore-att-paths` / `--no-ignore-att-paths` CLI flag to control this behavior
- New `ignore_att_paths` parameter in the hashing API
- Updated documentation and tests

## Rationale

Per #586, attestation files should be signed independently to allow attestations to accumulate throughout a model's lifecycle without invalidating the original signature. This is especially important when using the `claims.jsonl` pattern where attestations are appended over time.

The exclusion is enabled by default but can be disabled for edge cases where users need different behavior.

Related: ossf/model-signing-spec#4

### DIFF
--- a/docs/demo.ipynb
+++ b/docs/demo.ipynb
@@ -1093,7 +1093,7 @@
       "outputs": [],
       "source": [
         "hashing_config = model_signing.hashing.Config().set_ignored_paths(\n",
-        "    paths=[\"README.md\"], ignore_git_paths=True\n",
+        "    paths=[\"README.md\"], ignore_git_paths=True, ignore_att_paths=True\n",
         ")"
       ]
     },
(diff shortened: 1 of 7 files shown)


mihaimaruseac:
Let's also discuss this PR on the next model signing SIG group as it fits with the work from #592 and #591.

ralphbean:
We talked in sig meeting today - this is blocked for the moment, waiting on a proposal from Mihai on how we should handle exclusions more generally (rather than hacking on more exclusion options).

Issue #587, opened 2025-12-10
How do we recommend that OMS signatures and other attestations be named in a model repository?


ralphbean:
## Model Transfer and Attestation Lifecycle

Models get moved around. How should we handle their supply chain security artifacts?

Let’s follow a model through its original training, to huggingface, to consumption by a corporate user and its path through downstream systems.

For each, we’ll provide two depictions of the model file layout. One using standalone sigstore bundles and another using a unified [JSON lines](https://jsonlines.org/) file.

Recall, in both cases, an individual attestation is always a sigstore bundle. In the standalone sigstore bundles model, each sigstore bundle is its own file. In the unified `.jsonl` model, each *line* of that file is a sigstore bundle. In both cases, every sigstore bundle contains an in-toto statement where the `subject` is the root hash of the model which can be computed with OMS tooling (see also https://github.com/sigstore/model-transparency/issues/565) and the `predicate` is either the OMS signature predicate, one of the [well known in-toto predicates](https://github.com/in-toto/attestation/tree/main/spec/predicates), or a custom predicate bespoke to the producer.

**Proposal:** we should encourage the **Unified Bundle Layout** which uses the unified `claims.jsonl` file where new signed attestations are appended.

### Origin

A model may be originally produced on a workstation or in a pipeline. There, it can be signed by its producer generating an OMS signature sigstore bundle.

```
Individual Files Layout                 Unified Bundle Layout
──────────────────────────────    ────────────────────────────────────
origin-workspace/                       origin-workspace/
├── model.safetensors                   ├── model.safetensors
├── config.json                         ├── config.json
└── model.sig                           └── claims.jsonl
     # OMS signature                          # Contains:
     # (signed by author@example.com)         #   - OMS signature (signed by author@example.com)
```

### SBOM and SLSA

The producer may also generate an SBOM attestation and SLSA provenance attestation also as sigstore bundles.

```
Individual Files Layout                 Unified Bundle Layout
──────────────────────────────    ────────────────────────────────────
origin-workspace/                       origin-workspace/
├── model.safetensors                   ├── model.safetensors
├── config.json                         ├── config.json
├── model.sig                           └── claims.jsonl
├── model.slsa.sigstore.json                 # Appended with:
│    # NEW: SLSA provenance                   #   - SLSA provenance (signed by ci-builder@example.com)
│    # (signed by ci-builder@example.com)     #   - SPDX SBOM (signed by ci-builder@example.com)
└── model.spdx.sigstore.json
[...]

PR #591, opened 2025-12-17
Change default signature filename to claims.jsonl

ralphbean:
## Summary

- **Change default signature filename** from `model.sig` to `claims.jsonl` to align with OMS bundled attestation conventions (see #587)
- **Add backward-compatible signature format support** — both legacy `model.sig` (single JSON) and new `claims.jsonl` (JSONL) formats are supported during a deprecation period
- **New `SignatureFormat` enum and format detection** auto-detects file format based on extension and content
- **JSONL append semantics** — writing to `claims.jsonl` appends new claims; legacy `.sig` overwrites with a deprecation warning
- **Verification iterates claims newest-to-oldest**, succeeding on the first match
- Updated all documentation (README, demo notebook, format spec) to reflect the new default
- The previous default can still be used via the `--signature` CLI option

## Test plan

- [x] New `format_compat_test.py` covers format detection, deprecation warnings, and JSONL append behavior
- [ ] Verify `model_signing sign` produces `claims.jsonl` by default
- [ ] Verify `model_signing verify` reads both `model.sig` and `claims.jsonl`
- [ ] Verify `--signature model.sig` override still works

Relates to #587

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/model_signing/_cli.py
+++ b/src/model_signing/_cli.py
@@ -425,6 +425,7 @@ def _sign_sigstore(
         )
         span.set_attribute("sigstore.use_staging", use_staging)
         try:
+            existed = signature.exists()
             ignored = _resolve_ignore_paths(
                 model_path, list(ignore_paths) + [signature]
             )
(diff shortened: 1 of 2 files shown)


mihaimaruseac:
Let's discuss these PRs (#592 and #591) on the next model signing SIG group. I'd like to have these merged before the next release, but we should make sure this is done in a backwards and forwards compatible way.

PR #592, opened 2025-12-17
Implement append functionality for claims.jsonl attestations

ralphbean:
Implements the unified bundle layout from issue #587, where attestations accumulate in a single `claims.jsonl` file as the model moves through its lifecycle.

## Changes

- Signature `write()` now appends to existing `claims.jsonl` files instead of overwriting
- Each signature is written as compact JSON on a single line (JSONL format)
- Signature `read()` reads the last line (most recent attestation) for backward compatibility  
- Updated both `sign_sigstore.py` and `sign_sigstore_pb.py` implementations
- Updated test helpers to handle JSONL format

## Test-Driven Development

This was implemented using RED-GREEN-REFACTOR:
- **RED**: Added `test_append_to_existing_claims_jsonl` that initially failed
- **GREEN**: Implemented append functionality to make the test pass
- **REFACTOR**: Updated existing tests to handle JSONL format

## Example Usage

```python
# First signing creates the file
model_signing.signing.sign("my-model", "claims.jsonl")

# Second signing appends to the file
model_signing.signing.sign("my-model", "claims.jsonl")

# File now contains two lines, each a complete sigstore bundle
```

This allows attestations to accumulate as models move through their lifecycle (training → registry → security review → production) without invalidating earlier signatures.

Builds on PR #591

Related to #587

### DIFF
--- a/src/model_signing/_signing/sign_sigstore.py
+++ b/src/model_signing/_signing/sign_sigstore.py
@@ -52,16 +52,14 @@ class Signature(signing.Signature):
     @override
     def write(self, path: pathlib.Path) -> None:
         # Convert to compact JSON (single line) for JSONL format
-        import json
-
-        bundle_dict = json.loads(self.bundle.to_json())
-        compact_json = json.dumps(bundle_dict, separators=(",", ":"))
+        # by removing newlines from the bundle's JSON output
+        bundle_json = self.bundle.to_json().replace("\n", "")
 
         # Append to file if it exists (for accumulating attestations)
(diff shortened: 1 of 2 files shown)


mihaimaruseac:
Let's discuss these PRs (#592 and #591) on the next model signing SIG group. I'd like to have these merged before the next release, but we should make sure this is done in a backwards and forwards compatible way.

ralphbean:
Dropping this. Will move everything in to #591.

PR #597, opened 2026-01-06
Add KMS signer support for AWS, GCP, Azure, and file-based backends

### LINKED ISSUE #203: Cutsom raw signers
I use the term "raw signer" as per the description in https://github.com/sigstore/model-transparency/issues/172

We'd like to have raw signer instantiations for:
- KMS signer
- HSM / PKCS11 signers

SequeI:
#### Summary

This commit adds support for signing models using Key Management Service (KMS) providers through KMS URIs. The implementation follows the same pattern as the existing PKCS11 signer, providing a consistent interface for raw signer instantiations. Only available provider for now is AWS.

- AWS KMS: `kms://aws/<key-id/key-arn>?region=<region>`

Closes #203 

Moved some testing resources and the check EC curve functions to not duplicate

```
❯ hatch run model_signing sign kms-key bert-base-uncased  \
  --kms_uri "kms://aws/arn:aws:kms:<>" \
  --signature model.sig
  
Found credentials in shared credentials file: ~/.aws/credentials
Signing succeeded

❯ hatch run model_signing verify key bert-base-uncased  \
  --public-key /aws-kms-pubkey.pem \
  --signature model.sig
  
Verification succeeded
```

##### Checklist

### DIFF
--- a/src/model_signing/_signing/sign_kms.py
+++ b/src/model_signing/_signing/sign_kms.py
@@ -1,0 +1,209 @@
+# Copyright 2025 The Sigstore Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import base64
+import hashlib
+from urllib.parse import parse_qs
+from urllib.parse import urlparse
+
+from cryptography.hazmat.primitives import hashes
+from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives.asymmetric import ec
+from google.protobuf import json_format
+from sigstore_models import intoto as intoto_pb
+from sigstore_models.bundle import v1 as bundle_pb
+from sigstore_models.common import v1 as common_pb
+from typing_extensions import override
+
+from model_signing._signing import sign_ec_key as ec_key
+from model_signing._signing import sign_sigstore_pb as sigstore_pb
+from model_signing._signing import signing
+from model_signing._signing.sign_ec_key import _check_supported_ec_key
+
+
+class AWSKMSBackend:
+    """AWS KMS backend for signing with keys stored in AWS KMS."""
+
+    def __init__(self, key_id: str, region: str | None = None):
+        """Initializes the AWS KMS backend.
+
+        Args:
+            key_id: The AWS KMS key ID or ARN.
+            region: Optional AWS region. If not provided, uses default region.
+        """
+        try:
+            import boto3
+        except ImportError as e:
+            raise RuntimeError(
+                "AWS KMS support requires 'boto3'. "
+                "Install with 'pip install boto3'."
+            ) from e
+
+        self._key_id = key_id
+        self._kms_client = boto3.client("kms", region_name=region)
+        self._public_key = self._get_public_key()
+
+    def _get_public_key(self) -> ec.EllipticCurvePublicKey:
+        response = self._kms_client.get_public_key(KeyId=self._key_id)
+        public_key_der = response["PublicKey"]
+        public_key = serialization.load_der_public_key(public_key_der)
+        _check_supported_ec_key(public_key)
+        return public_key
+
+    def sign(self, digest: bytes) -> bytes:
+        response = self._kms_client.sign(
+            KeyId=self._key_id,
+            Message=digest,
+            MessageType="DIGEST",
+            SigningAlgorithm=(
+                "ECDSA_SHA_256"
+                if self._public_key.curve.name == "secp256r1"
+                else "ECDSA_SHA_384"
+                if self._public_key.curve.name == "secp384r1"
+                else "ECDSA_SHA_512"
+            ),
+        )
+        sig_bytes = response["Signature"]
+        return sig_bytes
+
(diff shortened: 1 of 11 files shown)


mihaimaruseac:
+1, let's only add the support here when there is a significant need for this.

It's for the same reason why #148 is still open, but because we haven't yet seen an actual demand to hash models while they are on remote blob storage we did not spend cycles to implement it.

mihaimaruseac:
Confirmed that there is a usecase for this

SequeI:
@mihaimaruseac Which provider is the use case for?

IMO, I can revise this PR to add support for that specific provider first, then when need arises we can increase KMS provider options.

mihaimaruseac:
Yesterday at the OMS meeting @ralphbean said that there is a need for this PR

Hayden-IO:
But for all KMS providers? If you don't have a good way to test support across all of these ecosystems, I would advise against proactively adding all of them.

mihaimaruseac:
Oh, that's true. Waiting on @ralphbean to suggest which of the providers is important

ralphbean:
I only have a concrete use case for the AWS one.

SequeI:
Reworked PR for AWS provider only, pipelines failing due to `get dangerous oidc token` action, not related to PR. We should fix/rework this sometime if possible, I have notice a lot of failures from this action

@ralphbean Let me know if this is what you had in mind, I tested via full ARN and worked fine.

src/model_signing/_signing/sign_kms.py
```diff
@@ -0,0 +1,209 @@
...
+            raise ValueError(
+                "AWS KMS URI must have format: kms://aws/<key-id-or-arn>"
+            )
+        key_id = "/".join(path_parts)
+        if key_id.startswith("arn:aws:kms:"):
+            arn_parts = key_id.split(":")
+            if len(arn_parts) != 6 or arn_parts[5].split("/")[0] != "key":
+                raise ValueError(
+                    "AWS KMS ARN must have format: "
+                    "arn:aws:kms:<region>:<account-id>:key/<key-id>"
```
ralphbean, line 119: The ARN validation here rejects valid AWS KMS ARNs:

1. Keys in `aws-cn` (China) and `aws-us-gov` (GovCloud) partitions — e.g. `arn:aws-cn:kms:...` — are rejected because only `arn:aws:kms:` prefix is checked.
2. Alias-based ARNs (`arn:aws:kms:region:account:alias/my-alias`) are rejected because only `key/` resource type is accepted.
3. The `elif "/" in key_id` branch also rejects alias ARNs that don't match the `arn:aws:kms:` prefix.

Consider relaxing the validation here — AWS KMS will validate the key ID/ARN authoritatively when the API call is made, so client-side validation that rejects valid inputs does more harm than good.

src/model_signing/_signing/sign_kms.py
```diff
@@ -0,0 +1,209 @@
...
+                if self._public_key.curve.name == "secp384r1"
+                else "ECDSA_SHA_512"
+            ),
+        )
+        sig_bytes = response["Signature"]
+        return sig_bytes
+
+    def get_public_key(self) -> ec.EllipticCurvePublicKey:
+        return self._public_key
+
```
ralphbean, line 82: The ternary chain here silently maps any unrecognized curve to `ECDSA_SHA_512`. Consider using a dict lookup instead — it's more readable, explicitly lists the supported mappings, and raises `KeyError` naturally for unrecognized curves:

```python
_CURVE_TO_ALGORITHM = {
    "secp256r1": "ECDSA_SHA_256",
    "secp384r1": "ECDSA_SHA_384",
    "secp521r1": "ECDSA_SHA_512",
}
# then: SigningAlgorithm=_CURVE_TO_ALGORITHM[self._public_key.curve.name]
```

tests/_signing/kms_test.py
```diff
@@ -0,0 +1,64 @@
...
+            _parse_kms_uri("kms://azure/vault/key")
+
+        with pytest.raises(ValueError, match="Unsupported KMS provider"):
+            _parse_kms_uri("kms://file/path/to/key.pem")
+
+        with pytest.raises(ValueError, match="AWS KMS URI must be either"):
+            _parse_kms_uri("kms://aws/key-id/extra")
+
+        with pytest.raises(ValueError, match="AWS KMS ARN must have format"):
+            _parse_kms_uri("kms://aws/arn:aws:kms:invalid")
```
ralphbean, line 64: The tests here only cover URI parsing. There are no tests for the actual signing path — no mocked `boto3` tests for `Signer` instantiation, `sign()`, or round-trip sign-and-verify.

Every other signer in this project (key, certificate, sigstore, pkcs11) has sign-and-verify round-trip tests in `tests/api_test.py`. Consider adding at minimum a unit test with mocked `boto3` that verifies the `Signer` can produce a valid `Signature` object that can be verified with the existing EC key `Verifier`.

src/model_signing/_signing/sign_kms.py
```diff
@@ -0,0 +1,209 @@
...
+        public_key = serialization.load_der_public_key(public_key_der)
+        _check_supported_ec_key(public_key)
+        return public_key
+
+    def sign(self, digest: bytes) -> bytes:
+        response = self._kms_client.sign(
+            KeyId=self._key_id,
+            Message=digest,
+            MessageType="DIGEST",
+            SigningAlgorithm=(
```
ralphbean, line 69: After retrieving the public key, consider also checking `response["KeyUsage"]` to verify the key is configured for signing (`SIGN_VERIFY`). AWS KMS keys can have `ENCRYPT_DECRYPT` usage — if one of those is passed here, the `sign()` call will fail later with an opaque AWS API error. An early check like:

```python
if response.get("KeyUsage") \!= "SIGN_VERIFY":
    raise ValueError(
        f"KMS key must have KeyUsage=SIGN_VERIFY, got {response.get('KeyUsage')}"
    )
```

would give a clear, actionable error message.

PR #600, opened 2026-01-16
fix: standardize CLI flags to use hyphens

SequeI:
#### Summary

Use hyphens instead of underscores for all CLI flags (e.g., --trust-config instead of --trust_config). Underscore variants still accepted via token_normalize_func for backwards compatibility.

I noticed a bunch of inconsistencies, using hyphens strictly going forward would be most consistent imo but I do not mind if we standardize across underscores. Biggest issues found were the following inconsistent commands

1. `"--ignore-git-paths/--no-ignore-git-paths"` (inconsistent with the command below which is near identical)
2. `"--ignore_unsigned_files/--no-ignore_unsigned_files"` (hyphens mixed with underscores)
etc etc

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ All versions prior to 1.0.0 are untracked.
 -Added the `digest` subcommand to compute and print a model's digest. This enables other tools to easily pair the attestations with a model directory.
 
 ### Changed
-- ...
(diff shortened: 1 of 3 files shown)


mihaimaruseac (approved):
I was thinking that we'd have a breaking change, but great find on token normalization!

We can probably document that the next major release will only support the options using `-`, but that doesn't block this PR.

Thank you!

PR #626, opened 2026-03-31
add throughput reporting, warmup, stats, and CSV export to exp_hash

alekseevpavel04:
#### Summary

The hash benchmark only reported min time per run, making it hard to compare algorithms by actual throughput or spot variance across runs. Added throughput (MB/s) to the default output, a --stats flag for mean/stdev columns, a --warmup arg to discard cold-cache runs before timing, and --output to dump a CSV for further analysis.

All timing logic is now in _run_benchmark(), which also fixes a shadowing bug where the inner closure was named `hash`, overriding the builtin. A _print_summary() table at the end shows peak throughput per algorithm across all sizes, which is the number most people actually care about.

Run with --stats --output results.csv to get the full picture. Default behavior (just min time + throughput column) is backward compatible.

##### Checklist

### DIFF
--- a/benchmarks/exp_hash.py
+++ b/benchmarks/exp_hash.py
@@ -16,7 +16,12 @@
 """Script for running a benchmark to pick a hashing algorithm."""
 
 import argparse
+import csv
+import dataclasses
+import statistics
+import sys
 import timeit
+from pathlib import Path
 from typing import Final
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


mihaimaruseac:
You need to do the DCO for the commit.

alekseevpavel04:
Closing this since I'm not able to follow up on it any more. Thanks for taking a look!

PR #637, opened 2026-05-29
fix: scope hatch packages directive to wheel target only

### LINKED ISSUE #636: sdist installs produce empty wheel due to packages directive scope
**Description**

The `packages` directive in `pyproject.toml` is under `[tool.hatch.build]` (general), which applies to both wheel and sdist targets. Hatchling's sdist builder flattens the `src/` layout, placing `model_signing/` at the archive root, but the embedded `pyproject.toml` still references `packages = ["src/model_signing"]`. When pip installs from the sdist, hatchling finds no files at that path and produces an empty wheel with zero Python modules.

To reproduce:

```bash
rm -rf dist/ .venv-sdist-test
hatch build -t sdist
python -m venv .venv-sdist-test
.venv-sdist-test/bin/pip install dist/*.tar.gz --no-cache-dir --quiet
.venv-sdist-test/bin/python -c "import model_signing; print('OK')"
# ModuleNotFoundError: No module named 'model_signing'
rm -rf dist/ .venv-sdist-test
```

Fix: move the directive to `[tool.hatch.build.targets.wheel]` so the sdist target uses hatchling's default behavior (preserve the full source tree including `src/`).

**Version**

model-signing 1.1.1 (current main), hatchling 1.29.0

jonburdo:
The packages directive was under `[tool.hatch.build]` (general), which applies to both wheel and sdist targets. Hatchling's sdist builder flattens the src-layout, placing model_signing/ at the archive root, but the embedded pyproject.toml still references src/model_signing - a path that no longer exists in the sdist. This causes pip to install an empty wheel with zero Python modules.

Moving the directive to `[tool.hatch.build.targets.wheel]` lets the sdist target use hatchling's default behavior (preserve the full source tree including the src/ prefix), so the packages path matches the actual layout when installing from source distributions.

Resolves: #636

#### Summary

This allows installation from sdist. I've tested with a script:
`test-sdist.sh`
```bash
#!/bin/bash
# Test whether rh-model-signing installs correctly from sdist.
# Usage: ./test-sdist.sh <branch>
set -euo pipefail

branch=${1}

git checkout "$branch" --quiet
rm -rf dist/ .venv-sdist-test
hatch build -t sdist
python -m venv .venv-sdist-test
.venv-sdist-test/bin/pip install dist/*.tar.gz --no-cache-dir --quiet
.venv-sdist-test/bin/python -c "import model_signing; print('PASS')"
rm -rf dist/ .venv-sdist-test
```
This fails on the current `main` (https://github.com/sigstore/model-transparency/commit/7643e29333d1eb993115074a18467165ca09e80a):

```
$ ./test-sdist.sh main
──────────────────────────────────────────────────────────────────────────────────────────── sdist ─────────────────────────────────────────────────────────────────────────────────────────────
dist/model_signing-1.1.1.tar.gz

Traceback (most recent call last):
  File "<string>", line 1, in <module>
    import model_signing; print('PASS')
    ^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'model_signing'
```

But is fixed on this branch:
```
$ ./test-sdist.sh fix/sdist-build-config-sigstore
──────────────────────────────────────────────────────────────────────────────────────────── sdist ─────────────────────────────────────────────────────────────────────────────────────────────
dist/model_signing-1.1.1.tar.gz

PASS
```

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ All versions prior to 1.0.0 are untracked.
 - Standardized CLI flags to use hyphens (e.g., `--trust-config` instead of `--trust_config`). Underscore variants are still accepted for backwards compatibility via token normalization.
 
 ### Fixed
+- Fixed a bug where installing from the sdist produced an empty wheel with zero Python modules. The hatch `packages` directive was scoped to all build targets instead of the wheel target only, causing the sdist's flattened layout to not match the expected `src/` path. ([#636](https://github.com/sigstore/model-transparency/issues/636))
 - Fixed a bug where ignored symlinks could raise `ValueError`s if allow_symlinks was unset, even though they were skipped during serialization. ([#550](https://github.com/sigstore/model-transparency/pull/550))
 - Fixed a bug where any PEM encoded key could be read during the key-based flows which resulted in a Python exception because the rest of the code only supported elliptic curve keys. ([#573](https://github.com/sigstore/model-transparency/pull/573))
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,7 +75,7 @@ PyPI = "https://pypi.org/project/model-signing/"
 [tool.hatch.version]
 path = "src/model_signing/__init__.py"
 
-[tool.hatch.build]
+[tool.hatch.build.targets.wheel]
 packages = ["src/model_signing"]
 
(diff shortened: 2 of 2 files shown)


mihaimaruseac (approved):
Thank you

PR #639, opened 2026-06-09
tests: Use new testing identity token source

jku:
* New testing token is more reliably available and has a longer time-to-live
* Old token source will be deprecated later

See https://github.com/sigstore-conformance/extremely-dangerous-public-oidc-beacon/pull/64 for context. This PR was partially machine generated.


mihaimaruseac (approved):
Thank you!!

PR #647, opened 2026-07-15
reject signed manifest paths that escape the model directory

Support-itecz:
#### Summary

With `ignore_unsigned_files` enabled, `verifying.Config.verify` builds the list of files to hash by joining each resource identifier from the signed manifest onto the model path, with no check that the result stays inside the model directory. A manifest whose resource name is `../secret.txt` (or `../../etc/hosts`) makes the verifier open and hash a file outside the model root. Anyone whose signature your configuration accepts (for example a keyless identity you verify against) can therefore steer reads outside the intended directory during verification.

This resolves each joined path and rejects any that escape the model root before hashing, matching the containment approach already used for ignore-paths in `_cli._resolve_ignore_paths`. Keeping the check in the verifier means every signing method benefits without callers needing their own guard. Added a key-signed regression test that crafts a manifest with a `..` path and confirms verification now raises, plus a positive case that in-model paths still verify. To reproduce before the fix: sign a manifest containing a `../` resource, then `verify(..., ignore_unsigned_files=True)` reads the outside file and can pass.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All versions prior to 1.0.0 are untracked.
 - Standardized CLI flags to use hyphens (e.g., `--trust-config` instead of `--trust_config`). Underscore variants are still accepted for backwards compatibility via token normalization.
 
 ### Fixed
+- Fixed a path traversal during verification with `ignore_unsigned_files`, where a signed manifest whose resource paths contained `..` caused files outside the model directory to be read and hashed. Such paths are now rejected.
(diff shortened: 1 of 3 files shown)


stefanberger:
You should put your description above into the patch description.

Support-itecz:
Done, moved the description into the commit message.

mihaimaruseac:
Let's wait on this one until we have all conformance tests in https://github.com/ossf/model-signing-spec and validate the spec there.

Support-itecz:
any update?

PR #648, opened 2026-07-16
reject signing certificates not marked for code signing

Support-itecz:
#### Summary

`_verify_certificates` decides a signing certificate is usable when it has the digitalSignature key usage bit or a code-signing extended key usage, treating the two as alternatives. An extended key usage is restrictive when present (RFC 5280 4.2.1.12), so a certificate whose only EKU is `serverAuth` (an ordinary TLS certificate) is still accepted as long as digitalSignature is set. Because the default trust roots are the certifi bundle, any holder of a publicly trusted TLS certificate can produce a signature this verifier accepts.

The fix rejects a certificate whose extended key usage lists neither code signing nor anyExtendedKeyUsage, even when the key usage bit is set. Keeping the check inside the verifier covers every certificate-based verification path without callers validating the chain themselves. Added a regression test that mints a `serverAuth`-only leaf under a private root and confirms it is now rejected, along with positive cases for a code-signing leaf and a leaf with no EKU.

To reproduce before the fix: build a leaf with `ExtendedKeyUsage([serverAuth])` and `KeyUsage(digital_signature=True)` chaining to a root in the verifier's store, then verify a signature made with its key. Verification passes.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All versions prior to 1.0.0 are untracked.
 - Standardized CLI flags to use hyphens (e.g., `--trust-config` instead of `--trust_config`). Underscore variants are still accepted for backwards compatibility via token normalization.
 
 ### Fixed
+- Fixed certificate verification accepting a signing certificate whose extended key usage does not permit code signing (for example a TLS `serverAuth` certificate), as long as the digitalSignature key usage bit was set. ([#648](https://github.com/sigstore/model-transparency/pull/648))
 - Fixed a bug where reusing a single `verifying.Config` across models let the ignore paths and guessed hashing configuration from one verification carry over into later ones. A file that a later model's signature never excluded could be silently skipped instead of reported as unsigned. ([#650](https://github.com/sigstore/model-transparency/pull/650))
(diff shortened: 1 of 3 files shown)


mihaimaruseac:
This looks good, but I will want to run some tests against existing signed models, to make sure they can still be verified.

Meanwhile, can you solve the conflicts please?

Support-itecz:
Conflicts resolved, rebased onto main. Main picked up its own tests/_signing/certificate_test.py from #649, so I folded the EKU tests into that file instead of keeping two copies; full suite passes locally (203 tests).

On the compatibility testing: certs with a codeSigning EKU or with no EKU at all still verify unchanged. The only signatures newly rejected are ones whose cert carries an EKU listing neither codeSigning nor anyExtendedKeyUsage, which is the case the fix targets.

Support-itecz:
Conflicts resolved. Rebased on main and folded my EKU tests into the new tests/_signing/certificate_test.py alongside the SAN pinning ones; full unit suite passes (200). On testing against existing signed models: certificates with no EKU at all are still accepted, only an EKU listing neither codeSigning nor anyExtendedKeyUsage is rejected, and Fulcio-issued leaves carry codeSigning, so already-signed models should verify unchanged.

Support-itecz:
any update?

mihaimaruseac (approved):

mihaimaruseac:
Thank you!

PR #650, opened 2026-07-18
scope hashing config per verify call to stop ignore-path leak

Support-itecz:
#### Summary

`verifying.Config.verify` carried hashing state across models when the config is reused (the module docstring recommends reusing one `Config` for many models):
- the guessed hashing config was cached on the instance, so a later model was hashed with an earlier model's serialization settings
- each signed manifest's ignore paths were added to that shared config via `add_ignored_paths` and never reset, so they accumulated across `verify()` calls
- an extra unsigned file placed at a path some earlier model ignored was skipped instead of reported during a later model's verification, so a tampered model verified clean

Build the hashing config per call (deepcopy a caller-supplied one, else a fresh guess) and apply the signed ignore paths to that local copy so nothing persists across calls. The regression test signs one model that ignores a file, then a second model whose signature excludes nothing, and confirms an unsigned file planted at the first model's ignored path is now reported.

##### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All versions prior to 1.0.0 are untracked.
 - Standardized CLI flags to use hyphens (e.g., `--trust-config` instead of `--trust_config`). Underscore variants are still accepted for backwards compatibility via token normalization.
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


mihaimaruseac (approved):

mihaimaruseac:
Thank you and nice bug catch!

PR #652, opened 2026-07-29
raise instead of skipping model directories that cannot be listed

Support-itecz:
#### Summary

`check_file_or_directory` accepts any path that stats as a directory, but `glob("**/*")` drops a subtree it cannot list without surfacing the error.

- CPython swallows the failure inside the per-directory generator (`PermissionError` on 3.10-3.12, `OSError` in `glob._Globber` on 3.13+)
- the manifest records only files, so an unreadable directory contributes zero entries and nothing raises
- `mkdir model/evil; cp backdoor.py model/evil/; chmod 0111 model/evil` keeps `backdoor.py` out of the manifest and `verify` still reports success; mode 0111 still grants search, so whatever loads the model opens the planted file normally
- both the file and shard serializers call this helper, so one check covers both walk loops

The docstring already said this call raises when there are permission errors. Signing or verifying a tree you cannot fully read now fails instead of quietly covering less than it appears to.

##### Checklist


Support-itecz:
rebased onto main to clear the CHANGELOG conflict, CI is green. any update?

PR #653, opened 2026-07-30
record zero-length files in the shard serialization manifest

Support-itecz:
#### Summary

`_get_shards` guards its loop with `if path_size > 0`, so a zero-length file produces no shards and never becomes a manifest item. Shard serialization leaves every empty file in the model out of the signature, even though the file was there when the model was signed.

The verifier re-serializes the model the same way, so the omission is symmetric and nothing reports a difference. Against a model signed with `use_shard_serialization()`:

- deleting a signed empty file verifies clean
- adding new empty files verifies clean, including one that creates a new directory (`sub/__init__.py`)
- with `ignore_unsigned_files`, a signed-as-empty file can be filled with content and still verifies clean

Empty marker files decide real loader behavior (`__init__.py`, `py.typed`, feature-flag files), so this is a gap in what the signature covers rather than a cosmetic difference. File serialization already recorded empty files, which is why `file_test.py` asserts `test_folder_model_empty_file_gets_included` while the shard suite asserted the opposite and the shard goldens for both empty-file fixtures were zero-byte. Dropping the guard makes `_endpoints(shard_size, 0)` yield one `(path, 0, 0)` shard, and `ShardedFileHasher.set_shard` accepts `end == start` for it, so the two serializers agree.

Shard signatures over models that contain empty files have to be regenerated. The unit suite passes (198), and the version-compat verify scripts under `scripts/tests` still pass for v0.2.0 through v1.1.0, since those signatures are file-serialized.

##### Checklist


Support-itecz:
Rebased onto main to clear the CHANGELOG conflict, unit suite still passes (201). Any update?

